PR #2, opened 2017-10-31
initial upload

ambrosejcarr:
Initial upload of the `scsequtils` class, converted and uploaded to humancellatlas as `sctools`

I put together a basic readme of the main package classes and their use, as well as the command line argument(s) (currently only one) that I'm using for optimus. 

The main functions are in the `sctools` package, and tests for each module are in `sctools/test`

There are quite a few testing files whose size I have tried to keep small; there is no need to review any code in `sctools/test/data/`


setup.py
```diff
@@ -0,0 +1,41 @@
+from setuptools import setup
+
+CLASSIFIERS = [
+    "Development Status :: 4 - Beta",
+    "Natural Language :: English",
+    "License :: OSI Approved :: GNU General Public License v2 (GPLv2)",
```
dshiga: The license checked into the repo is MIT. Other Mint repos are BSD-3 so would prefer to use that for consistency, unless there's a good reason not to.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+        """
+        :param str bam_file: location of bam file
+        """
+        if not isinstance(bam_file, str):
+            raise TypeError('bam_file must be type str, not %s' % type(bam_file))
+        self.bam_file = bam_file
+
+    def tag(self, output_bam_name, tag_generators):
+        """
+        given a bam file and tag generators in the same order, adds tags to the .bam file,
```
dshiga: what do you mean by "in the same order" here? same order as something in the bam file?

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+
+        :param str output_bam_name: name of output tagged bam.
+        :param [fastq.TagGenerator] tag_generators: generators that yield Tag objects
+          (see fastq.Tag)
+        """
+        inbam = pysam.AlignmentFile(self.bam_file, 'rb', check_sq=False)
+        outbam = pysam.AlignmentFile(output_bam_name, 'wb', header=inbam.header)
+        try:
+            # zip up all the iterators
+            for *tag_sets, sam_record in zip(*tag_generators, inbam):
```
dshiga: I'm not familiar with pysam. I'm assuming this does not pull all reads in the bam into memory at once, although naively that's that it looks like. Can you confirm?

ambrosejcarr: Confirmed; it's an iterator, and pysam is the python wrapper for htslib. 

This line is zipping up three iterators that are lazily read. You get 4 fastq lines and one sam line per iteration.

src/sctools/bam.py
ambrosejcarr: Correct. All files need equivalent sort order. I'll make the doc string clearer.

setup.py
ambrosejcarr: No reason. I'll switch it.

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+
+def zip_readers(*readers, indices=None):
+    """zip together multiple fastq objects, yielding records simultaneously.
+
+    :param [Reader] readers:
+    :param set indices: set of indices to iterate over
+
+    :return Iterator: iterator over tuples of records, one from each passed Reader object.
+    """
+    # iterators = [iter(r) for r in readers]
```
dshiga: Is this commented out line needed?

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+                yield record
+                indices.remove(idx)
+
+                # stopping condition
+                if not indices:
+                    break
+
+
+def zip_readers(*readers, indices=None):
+    """zip together multiple fastq objects, yielding records simultaneously.
```
dshiga: This is actually a more general function, right? It doesn't have to be a fastq object, just a Reader.

ambrosejcarr: Good point. I'll edit the docstring!

src/sctools/reader.py
ambrosejcarr: Nope, will remove.

setup.py
```diff
@@ -0,0 +1,41 @@
...
+    url='https://github.com/humancellatlas/scsequtil.git',
+    author='Ambrose J. Carr',
+    author_email='mail@ambrosejcarr.com',
+    package_dir={'': 'src'},
+    packages=['sctools', 'sctools/test'],
+    test_suite='setup.sctools_test_suite',  # runs tests twice for who knows what reason
+    install_requires=[
+        'pysam',
+        'numpy',
+        'google-cloud',
```
dshiga: This brings in a huge web of dependencies. We were doing the same in secondary-analysis but changed it to install a more targeted set of google libraries. (We ran into an issue with one of the many packages that google-cloud installs, which it turned out we didn't actually need.)

ambrosejcarr: Ah, very cool. Do you think it would be adequate to pull the same set of libraries you're using in the secondary-analysis repo?

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+    @property
+    def filenames(self):
+        return self._files
+
+    def __len__(self):
+        """
+        return the length of the Reader object. this should typically not be used with sys.stdin,
+        as it will consume the input
+        """
+        return sum(1 for _ in self)
```
dshiga: Would be helpful to clarify that this gives the sum of line counts for all files in Reader and requires fully reading all files.

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def __iter__(self):
+        for file_ in self._files:
+
+            # set correct mode
+            if self._mode == 'r' and any(file_.endswith(s) for s in ['.gz', '.bz2']):
```
dshiga: Why do we treat this as a special case? We already set mode = 'r' when self._mode == 'r' by default in the else. In this special case we set it to 'rt' instead. And according to this, r and rt are the same thing:
https://stackoverflow.com/questions/23051062/open-files-in-rt-and-wt-modes

ambrosejcarr: the `gzip` and `bzip2` libraries _do_ make a distinction between 'r' and 'rt'. It's very annoying: 

https://docs.python.org/3/library/gzip.html#gzip.open

ambrosejcarr: ... if they have proper extensions. I will look into a better way of detecting compressing type.

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+                    yield record
+            finally:  # clean up
+                f.close()
+
+    @property
+    def size(self):
+        """return the collective size of all files being read in bytes"""
+        return sum(os.stat(f).st_size for f in self._files)
+
+    def select_indices(self, indices):
```
dshiga: Indices is a set of line numbers that we want to filter down to, right? Or is it more general than that? If it's line numbers it might be clearer to call this select_lines.

setup.py
dshiga: Not sure, depends which actual google modules you're using. Are you using it just for reading from buckets? In that case you probably just need google-cloud-storage.

src/sctools/reader.py
ambrosejcarr: `select_record_indices` maybe? It's not pulling out lines/records themselves, just indices to them.

setup.py
ambrosejcarr: I'll verify, but I think that's all.

src/sctools/reader.py
dshiga: Okay, in that case I think what you have is perfectly reasonable here. Maybe just add a comment explaining that rt really does differ from r in this case.

src/sctools/gtf.py
```diff
@@ -0,0 +1,147 @@
...
+from . import reader
+
+
+class Record:
+    """
+    Simple namespace object that makes the fields of a GTF record available. Subclassed
+    to create records specific to exons, transcripts, and genes
+    """
+
+    __slots__ = ['_fields', '_attribute']
```
dshiga: Should this be _attributes plural? Looks like it's a dict of k/v pairs below.

src/sctools/gtf.py
```diff
@@ -0,0 +1,147 @@
...
+    _del_non_letters = ''.join(
+        set(string.printable).difference(string.ascii_letters))
+
+    def __init__(self, record):
+        """
+        :param str record: input record from file
+        """
+        fields = record.strip(';\n').split('\t')
+        self._fields = fields[:8]
+        self._attribute = {
```
dshiga: Worth a comment explaining the parsing that's happening here, and/or a short example of what fields[8] looks like.

src/sctools/gtf.py
```diff
@@ -0,0 +1,147 @@
...
+        else:
+            return size
+
+    def get_attribute(self, key):
+        """access an item from the attribute field of a GTF file.
+
+        :param key: item to access
+        :return str: value of item
+        """
+        try:
```
dshiga: you could replace 94-97 with `return self._attribute.get(key)`

src/sctools/gtf.py
```diff
@@ -0,0 +1,147 @@
...
+
+    @property
+    def frame(self):
+        return self._fields[7]
+
+    @property
+    def size(self):
+        size = self.end - self.start
+        if size < 0:
+            raise ValueError('invalid record: negative size %d (start > end)' % size)
```
dshiga: Why not calculate this in __init__ and throw an error there? That would protect against creation of invalid records. Or is there value in being able to parse invalid ones?

src/sctools/gtf.py
```diff
@@ -0,0 +1,147 @@
...
+        :param list|str files: (default sys.stdin) file or list of files to be read.
+        :param mode: (Default 'r') returns string objects. Change to 'rb' to
+          return bytes.
+        :param str|bytes header_comment_char: (default #) character that marks headers, which are
+          ignored
+        """
+        super().__init__(files, mode, header_comment_char)  # has different default args from super
+
+    def __iter__(self):
+        for line in super().__iter__():
```
dshiga: Would `for line in super():` accomplish the same thing?

dshiga: Actually I take that back, I like being explicit here.

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+    :property average_quality: return the mean quality of FastqRecord
+    """
+    __slots__ = ['_data']
+
+    def __init__(self, record):
+        """
+        :param [str|bytes] record: list of four strings or bytes objects defining a
+          single fastq record
+        """
+        self._data = list(record)
```
dshiga: Is there a special reason why we are making a new list rather than using `record` as is, or are you just being careful? If there's a special reason, would be worth a comment.

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @property
+    def name(self):
+        return self._data[0]
+
+    @name.setter
+    def name(self, value):
+        if not isinstance(value, (bytes, str)):
```
dshiga: Currently you can create a record with an invalid name type but setting name to an invalid type after the fact throws an error. Would it be worth having `__init__` call the setters so that this error is also thrown at record creation time? Or is there a reason why you would want to allow invalid records to be constructed?

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            self._data[1] = value
+
+    @property
+    def name2(self):
+        return self._data[2]
+
+    @name2.setter
+    def name2(self, value):
+        if not isinstance(value, (bytes, str)):
```
dshiga: Good to have these constraints. In addition to restricting types, are there other constraints that would be worth adding? Do you frequently encounter fields with mangled format that you would want to reject for example?

src/sctools/gtf.py
ambrosejcarr: Actually it's more about limiting unnecessary processing of records; I wanted to maximize speed here and so do no parsing of the records until the data is specifically requested. 

Often only a few fields of the record are of interest, or a particular type of record. For example, often times on gene records are wanted, and therefore there is no value in processing the much more frequent exon or transcript records. 

I grudgingly added the variable field parsing to `init` (`field[8:]`, which you correctly wanted an explanation for) because it contains gene id information that is normally of interest.

src/sctools/fastq.py
ambrosejcarr: My docstring in line 21 is inaccurate; it should say "iterable" instead of list. This will often be read as a tuple. However, I require the ability to change the record, and so need it to be mutable. Hence the list conversion.

src/sctools/fastq.py
ambrosejcarr: I can't think of a value for invalid records. I don't think calling the setters at `__init__` time will introduce a significant overhead, so I'll switch to this. Thanks for the suggestion. 👍

src/sctools/fastq.py
ambrosejcarr: Yea; technically they are supposed to start with '@'. I can add this in.

ambrosejcarr: That said, I want to be careful to balance validation with speed; all quality strings should contain only ASCII letters and the name field has a specific structure defined by the sequencer that generates the fastq file. However, these are expensive to validate and might be better served by external validation tools (or expanded later, if we find we want to apply tools that require the structures of those fields to be intact/correct"

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+
+    @quality.setter
+    def quality(self, value):
+        if not isinstance(value, (bytes, str)):
+            raise TypeError('quality must be str or bytes')
+        else:
+            self._data[3] = value
+
+    def __bytes__(self):
+        try:
```
dshiga: Do you expect _data to usually be in bytes, and/or do you prefer to optimize speed for the bytes case, rather than for string? That appears to be the choice you're making here, just want to confirm.

dshiga: Would it be worth moving implementation of __bytes__ and __str__ into the BytesRecord and StrRecord classes, so you don't have to try catch?

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+    :property bytes name2: second name field
+    :property bytes quality: quality field
+
+    Also defines several methods for accessing SEQC annotation fields:
+    :property float average_quality: return the mean quality of FastqRecord
+    """
+
+    def average_quality(self):
+        """return the average quality of this record"""
+        return sum(c for c in self.quality[:-1]) / (len(self.quality[:-1]) - 1) - 33
```
dshiga: Why minus 33? A comment might be helpful here.

src/sctools/fastq.py
ambrosejcarr: I expect it to almost always be bytes (and it will always be bytes for our pipelines), as
it's faster to convert from bytes to 2-bit encodings; which you will see later in the package. 

I used to have these inside the respective classes as you're suggesting. The reason I moved away from it was essentially just to reduce the amount of code in the package. My logic was that anyone using strings no longer cares about speed, and therefore it's OK to add small inefficiencies for string records. 

I am very willing to move it back if you feel that would be better!

src/sctools/fastq.py
ambrosejcarr: Absolutely correct. It's because of an old solexa/illumina conversion.

src/sctools/fastq.py
ambrosejcarr: .. I will also make a note to be clear bytes is the preferred implementation.

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+            yield record_type(record)
+
+    def estimate_sequence_length(self):
+        """
+        estimate the sequence length of a fastq file
+
+        :return int: mean
+        :return float: standard deviation
+        """
+        pass  # implement
```
dshiga: Maybe better to return NotImplementedError here like you do elsewhere

src/sctools/reader.py
```diff
@@ -0,0 +1,122 @@
...
+    """zip together multiple fastq objects, yielding records simultaneously.
+
+    :param [Reader] readers:
+    :param set indices: set of indices to iterate over
+
+    :return Iterator: iterator over tuples of records, one from each passed Reader object.
+    """
+    # iterators = [iter(r) for r in readers]
+    if indices:
+        iterators = zip(*[r.select_indices(indices) for r in readers])
```
mckinsel: Do you want to use the generator syntax here? I'm not actually sure if it makes a difference.

src/sctools/fastq.py
ambrosejcarr: I will remove it. Thanks.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+        :return [int]: list of unaligned indices, only returned if include_other is not zero.
+        """
+
+        # check chromosome
+        valid_chromosomes = [str(i) for i in range(1, 23)] + ['M', 'X', 'Y']
+        if isinstance(chromosome, int) and chromosome < 23:
+            chromosome = str(chromosome)  # try to convert
+        if chromosome not in valid_chromosomes:
+            raise ValueError('chromsome %s not valid. Must be one of %r' %
+                             (chromosome, valid_chromosomes))
```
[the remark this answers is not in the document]
ambrosejcarr: No; I should expand it to accept . Thanks for the comment 👍

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+        with pysam.AlignmentFile(self._file, self._open_mode) as fin:
+            specific, nonspecific = 0, 0  # counters
+            chromosome = str(chromosome)
+            chromosome_indices = []
+            other_indices = []
+
+            for i, record in enumerate(fin):
+
+                if not record.is_unmapped:  # record is mapped
+                    if chromosome in record.reference_name and specific < n_specific:
```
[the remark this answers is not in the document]
ambrosejcarr: I've only used this for chromosomes 19 and 21; I think you're right. I'll adjust it to use `==`. 

I was, ironically, trying to support the `chr19` vs `19` problem you identify above but there are ways to get that without this bug :)

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+                        chromosome_indices.append(i)
+                        specific += 1
+                    elif nonspecific < include_other:
```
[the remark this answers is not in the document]
ambrosejcarr: Not intentional, thanks for pointing this out.

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+    """Abstract base class for sets of barcodes, encoded in 2bit representation for fast operations
+
+    :method summarize_hamming_distances:  returns descriptive statistics on hamming distances
+      between pairs of barcodes
+
+    """
+
+    def __init__(self, barcode_set, barcode_length):
+        """
+        :param Counter barcode_set: dictionary
```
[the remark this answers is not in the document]
ambrosejcarr: True; I'll make this.. `barcodes`

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+            distances.append(bin(a ^ b).count('1'))
+        distances.sort()
+
+        return {
+            'minimum': distances[0],
+            '25th percentile': distances[int(len(distances) * .25)],
+            'median': distances[int(len(distances) * .5)],
+            'average': sum(distances) / len(distances),
+            '75th percentile': distances[int(len(distances) * .75)],
+            'maximum': distances[-1]
```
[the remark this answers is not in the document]
ambrosejcarr: I was trying to avoid using numpy in the package but eventually capitulated for the entropy calculations. I'll switch it in.

src/sctools/encodings.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        map_ = {65: 0, 67: 1, 84: 2, 71: 3}  # A, C, T, G
+
+        def __getitem__(self, byte):
+            try:
+                return self.map_[byte]
+            except KeyError:
+                if byte != 78:
+                    raise
+                return random.randint(0, 4)
```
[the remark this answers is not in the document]
ambrosejcarr: In python 3 it is exclusive. 

You're right about the other codes. I guess there is no reason not to include the full IUPAC list; I have only ever seen `N` ambiguous nucleotides in fastq sequence, which was the main purpose here. 

👍  for generalizability, however.

src/sctools/encodings.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Since no bases are encoded as 0, an empty triplet is interpreted as the end of the encoded
+    string; Three-bit encoding can be used to encode and decode strings without knowledge of their
+    length.
+    """
+
+    class ThreeBitEncodingMap:
+
+        # C: 1, A: 2, G: 3, T: 4, N: 6;  # note, not using 0
+        map_ = {65: 2, 97: 2, 67: 1, 99: 1, 71: 3, 103: 3, 84: 4, 116: 4, 78: 6, 110: 6}
```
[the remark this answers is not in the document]
ambrosejcarr: Yes; that's wonderful. I've been struggling with how to write this clearly.

src/sctools/reader.py
ambrosejcarr: Do you mean `iterators = zip(*(r.select_indices(indices) for r in readers))`?

I'll make sure it doesn't break, but yea; that would probably be more memory efficient.

src/sctools/encodings.py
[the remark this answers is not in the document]
mckinsel: I'm committed to python 2.7 forever, so I'm not an expert on this, but I think `randint` is inclusive in python3: https://docs.python.org/3/library/random.html#random.randint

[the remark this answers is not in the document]
ambrosejcarr: Oh I'm sorry, i was assuming I was using numpy here, which is exclusive. I even went and tested the numpy version >.<

You're completely correct. I'll change this.

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+    """
+    Fastq Reader, defines some special methods for reading and summarizing fastq data:
+
+    :method __iter__: Iterator over fastq Record objects
+    :method __len__: return number of records in file
+    :method estimate_sequence_length: estimate the length of fastq sequences in file
+    """
+
+    @staticmethod
+    def record_grouper(iterable):
```
dshiga: I understand this function better after talking in person. A comment about its purpose would probably be helpful, here or maybe in the class docstring.

src/sctools/reader.py
mckinsel: Yeah. I mean it probably has no real effect, but you're so careful about using iterators everywhere else.

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+            for tag in self.tags:
+                tags.extend(self.extract_tag(record, tag))
+            yield tags
+
+    @staticmethod
+    def extract_tag(record, tag):
+        """
+        :param FastqRecord record: record to extract from
+        :param Tag tag: defines tag to extract
+        :return tuple:
```
dshiga: Could use a more detailed comment about what's returned

src/sctools/reader.py
ambrosejcarr: No, I think this could really matter at scale. This is a good catch. 👍

src/sctools/fastq.py
```diff
@@ -0,0 +1,197 @@
...
+        estimate the sequence length of a fastq file
+
+        :return int: mean
+        :return float: standard deviation
+        """
+        pass  # implement
+
+
+# this could easily be expanded beyond fastq generated tags, we just don't have any use cases yet.
+Tag = namedtuple('Tag', ['start', 'end', 'sequence_tag', 'quality_tag'])
```
dshiga: I think I mostly understand now, but it was not clear to me at first what a fastq tag is and what TagGenerator was doing. More descriptive comments about what Tag and TagGenerator are for would be helpful.

ambrosejcarr: I feel like there is one too many levels of abstraction in this set of functions; this is what I was most concerned about being unclear. I'll put some time into improving the documentation and also maybe the code.

src/sctools/encodings.py
dshiga: Hmm, unless it would bog down performance, what about creating the map_ this way?
{ord('A'): 2, ord('a'): 2, ... }
That way you are documenting the mapping in code rather than a comment, and less chance for a typo to introduce an error.
Or you could have raw_map = {'A': 2, 'a': 2, ... } and then process that to create map_ = {65: 2, ... }

ambrosejcarr: Markus made the same (really good) suggestion. I'm going to change this; much much clearer the way you're both suggesting.

src/sctools/encodings.py
dshiga: This returns 0 to 4 inclusive, but I think you meant 0-3. `random.randrange(0, 4)` or `random.randint(0, 3)`

ambrosejcarr: Correct; I converted from numpy but forgot that numpy.randint and random.randint are different.

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+        :return np.array: barcode_length x 4 2d array
+        """
+        base_counts_by_position = np.zeros((self._barcode_length, 4), dtype=np.uint64)
+        keys = np.fromiter(self._data.keys(), dtype=np.uint64)
+        values = np.fromiter(self._data.values(), dtype=np.uint64)
+
+        for i in reversed(range(self._barcode_length)):
+            binary_base_representations, counts = np.unique(keys & 3, return_counts=True)
+            if weighted:  # todo weighted not working, values multiplication does not work
+                base_counts_by_position[i, binary_base_representations] = counts  # * values
```
dshiga: Maybe this should throw a NotImplementedError, or else take out this option altogether for now.

src/sctools/barcode.py
dshiga: Would be helpful to indicate that this is a dict of barcodes to counts. Do you need it to be a collections.Counter specifically, or any dict of barcodes to counts?

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+    :method from_whitelist: construct a PriorBarcodeSet from a whitelist file
+    :method summarize_hamming_distances:  returns descriptive statistics on hamming distances
+      between pairs of barcodes
+    """
+
+    @classmethod
+    def from_whitelist(cls, file_, barcode_length):
+        """Creates a barcode set from a whitelist file
+
+        :param str file_: location of the whitelist file. Should be formatted one barcode per line.
```
dshiga: It looks like the barcodes are expected to be strings like 'ACGT', not two bit or three bit encoded, etc. Would be helpful to specify that or give an example barcode string, since elsewhere in this package we expect two bit or three bit encoded strings.

src/sctools/barcode.py
ambrosejcarr: I think any mapping to counts is probably Ok; i'll verify and make explicit.

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+    @classmethod
+    def from_iterable_encoded(cls, iterable):
+        """construct an ObservedBarcodeSet from an iterable of encoded barcodes"""
+        return cls(Counter(iterable))
+
+    @classmethod
+    def from_iterable_strings(cls, iterable, barcode_length):
+        """construct an ObservedBarcodeSet from an iterable of string barcodes"""
+        tbe = TwoBit(barcode_length)
+        return cls(Counter(tbe.encode(b) for b in iterable))
```
dshiga: Don't you need to include barcode_length as a param to cls?

src/sctools/barcode.py
ambrosejcarr: it should throw a NotImplementedError or I can extract it to another open branch and add an issue. I think this will be useful, but we can drop it from this upload as I haven't figured out how to do this efficiently, yet.

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+    def __init__(self, barcode_set, barcode_length):
+        """
+        :param Counter barcode_set: dictionary
+        :param int barcode_length: the length of all barcodes in the set. Different-length barcodes
+          are not supported
+        """
+        if not isinstance(barcode_set, Mapping):
+            raise TypeError('barcode set must be a dict-like object mapping barcodes to counts')
+        self._data = barcode_set
+        self._barcode_length = barcode_length
```
dshiga: Should we throw an error if barcode_length is None?

src/sctools/barcode.py
```diff
@@ -0,0 +1,134 @@
...
+
+        :param bool weighted: (Default False) if True, each barcode is counted once for each time it
+          was observed
+
+        :return np.array[float]: array of size barcode_length containing floats in [0, 1]
+        """
+        return base4_entropy(self.base_frequency(weighted=weighted))
+
+
+class PriorBarcodeSet(BarcodeBase):
```
dshiga: I'm not clear on why two classes are needed here, rather than three factory methods that create instances of BarcodeBase from a file, from an iterable of strings, and an iterable of bit encoded strings. Once you've constructed them, they share all of their code in BarcodeBase.

src/sctools/platform.py
```diff
@@ -0,0 +1,71 @@
...
+                help='read 1 fastq record for a 10x genomics experiment')
+            parser.add_argument(
+                '--i1', required=True, help='i7 fastq record for a 10x genomics experiment')
+            parser.add_argument(
+                '--u2', required=True,
+                help='unaligned read-2 bam containing genomic information. Can be converted'
+                     'using picard FastqToSam')
+            parser.add_argument('-o', '--output-bamfile', required=True,
+                                help='filename for tagged bam')
+            args = vars(parser.parse_args())
```
dshiga: why convert to a dict here? Isn't it easier to use args.u2 etc below instead of args['u2']?

src/sctools/platform.py
```diff
@@ -0,0 +1,71 @@
...
+    molecule_barcode = fastq.Tag(start=16, end=24, quality_tag='UY', sequence_tag='UR')
+    sample_barcode = fastq.Tag(start=0, end=8, quality_tag='SY', sequence_tag='SR')
+
+    _tags = {
+        'r1': (cell_barcode, molecule_barcode),
+        'i1': (sample_barcode,)
+    }
+
+    @classmethod
+    def get_tag(cls, sequencing_read):
```
dshiga: Should this be get_tags plural? Returns a tuple of tags.

src/sctools/platform.py
```diff
@@ -0,0 +1,71 @@
...
+        """Extracts tags from fastq files_with_tags, attaches them to input_bamfile_name and writes
+        the result to output_bamfile_name
+
+        :param str input_bamfile_name: input bam
+        :param str output_bamfile_name: output bam
+        :param dict files_with_tags: dict mapping {filenames: tags}
+        """
+        tag_generators = []
+        for k, v in files_with_tags.items():
+            tag_generators.append(fastq.TagGenerator(cls.get_tag(k), files=v))
```
dshiga: The docstring says the key is the filename and the value is the tag, opposite of what is happening here.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+                open_mode = 'rb'
+            elif alignment_file.endswith('.sam'):
+                open_mode = 'r'
+            else:
+                raise ValueError('could not autodetect file type for alignment_file %s '
+                                 '(detectible suffixes: .sam, .bam)' % alignment_file)
+        self._file = alignment_file
+        self._open_mode = open_mode
+
+    def indices_by_chromosome(self, n_specific, chromosome, include_other=0):
```
dshiga: Maybe call it include_unaligned instead of include_other. Would it make sense to interpret -1 to mean include all unaligned reads? Is that a useful thing to be able to do?

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+    def tag(self, output_bam_name, tag_generators):
+        """
+        given a bam file and tag generators in the same order, adds tags to the .bam file,
+        writes the resulting file to output_bam_name.
+
+        :param str output_bam_name: name of output tagged bam.
+        :param [fastq.TagGenerator] tag_generators: generators that yield Tag objects
+          (see fastq.Tag)
+        """
+        inbam = pysam.AlignmentFile(self.bam_file, 'rb', check_sq=False)
```
dshiga: How about using `with pysam.AlignmentFile` the way you do on line 57?

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+        valid_chromosomes = [str(i) for i in range(1, 23)] + ['M', 'X', 'Y']
+        if isinstance(chromosome, int) and chromosome < 23:
+            chromosome = str(chromosome)  # try to convert
+        if chromosome not in valid_chromosomes:
+            raise ValueError('chromsome %s not valid. Must be one of %r' %
+                             (chromosome, valid_chromosomes))
+
+        with pysam.AlignmentFile(self._file, self._open_mode) as fin:
+            specific, nonspecific = 0, 0  # counters
+            chromosome = str(chromosome)
```
dshiga: You already converted to string on 52.

src/sctools/barcode.py
ambrosejcarr: Good idea. I might have some questions about implementation here, but I'll make this change.

src/sctools/bam.py
dshiga: Oh I see, this is not just unaligned reads but any read that does not match the chromosome we asked for.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+            chromosome_indices = []
+            other_indices = []
+
+            for i, record in enumerate(fin):
```
dshiga: It feels like some of the contents here, especially the if else block would be worth moving out into its own function that could be unit tested on its own.

src/sctools/bam.py
ambrosejcarr: i'll find a way to clarify; sounds like it produced some confusion.

ambrosejcarr:
Thanks very much @dshiga. I'll begin addressing your comments shortly!

src/sctools/fastq.py
dshiga: It's such a small amount of code and you have to have it somewhere - seems a little clearer to put it in BytesRecord and StrRecord.

src/sctools/bam.py
ambrosejcarr: This is a bit different; here I'm being lenient in case the user passed an int as chromosome. On 52 I was building up the set of allowable chromosomes.

src/sctools/platform.py
ambrosejcarr: For testing; it's hard to create namespace objects but easy to create keyword dictionaries; I'll make it clear that args can be passed that way in the docstring.

src/sctools/fastq.py
ambrosejcarr: Adjusted to make bytes and str both 1st class. StrRecord now subclasses Record, which operates on bytes. This should also indicate to the user (subtly) that bytes is the preferred implementation.

src/sctools/barcode.py
ambrosejcarr: Added comment in file_ docstring.

src/sctools/bam.py
ambrosejcarr: Adjusted docstring: 

"""Return the list of first n_specific indices of reads aligned to selected chromosome.

        If desired, will also return non-specific indices in a second list (can serve as negative
        control reads).

        :param int n_specific: number of aligned reads to return indices for
        :param str chromosome: only reads from this chromosome are considered valid
        :param int include_other: optional, (default=0), the number of reads to include that are
          NOT aligned to chromosome (could be aligned or unaligned read)
        :return [int]: list of indices to reads aligning to chromosome
        :return [int]: list of indices to reads NOT aligning to chromosome, only returned if
          include_other is not 0.
        """

src/sctools/bam.py
ambrosejcarr: I simplified the code. Since each list hashes its length, I don't need the counters; I can just test the length of `chromosome_indices` and `other_indices` directly. 

However, removing this into its own function would only serve to separate the type/value checking from the index extraction; I think this belongs where it is, but hope that the code simplification makes it clearer and reduces your desire to extract the code.

src/sctools/gtf.py
ambrosejcarr: Added explanation in docstring.

src/sctools/gtf.py
ambrosejcarr: left unchanged.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+        :param str chromosome: only reads from this chromosome are considered valid
+        :param int include_other: optional, (default=0), the number of reads to include that are
+          NOT aligned to chromosome (could be aligned or unaligned read)
+        :return [int]: list of indices to reads aligning to chromosome
+        :return [int]: list of indices to reads NOT aligning to chromosome, only returned if
+          include_other is not 0.
+        """
+
+        # acceptable chromosomes
+        valid_chromosomes = [str(i) for i in range(1, 23)] + ['M', 'X', 'Y']
```
mckinsel: I believe "MT" is also used to represent mitochrondia too.

src/sctools/bam.py
```diff
@@ -0,0 +1,125 @@
...
+                elif len(other_indices) < include_other:  # record is not mapped
+                    other_indices.append(i)
+
+                # check termination condition (we have the requisite number of reads
+                if len(chromosome_indices) == n_specific and len(other_indices) == include_other:
+                    break
+
+        if len(chromosome_indices) < n_specific or len(other_indices) < include_other:
+            warnings.warn('Only %d unaligned and %d reads aligned to chromosome %s were found in' 
+                          '%s' % (len(chromosome_indices), len(other_indices),
```
mckinsel: I think you've swapped `chromosome_indices` and `other_indices`.

src/sctools/encodings.py
```diff
@@ -0,0 +1,169 @@
...
+        nucleotides e.g. N
+        """
+        map_ = {ord('A'): 0, ord('C'): 1, ord('T'): 2, ord('G'): 3,
+                ord('a'): 0, ord('c'): 1, ord('t'): 2, ord('g'): 3}
+
+        def __getitem__(self, byte):
+            try:
+                return self.map_[byte]
+            except KeyError:
+                if byte != 78:
```
mckinsel: Just to keep everything consistent, you might want `byte in (ord("N"), ord("n"))`

src/sctools/barcode.py
```diff
@@ -0,0 +1,110 @@
...
+    def __iter__(self):
+        return iter(self._data)
+
+    def __len__(self):
+        return len(self._data)
+
+    def __getitem__(self, item):
+        return self._data[item]
+
+    def summarize_hamming_distances(self):
```
mckinsel: Could you add a little test of this method?

ambrosejcarr:
Marcus' most recent comment revealed some non-functional untested code. I ran a coverage report and added tests to missing modules. If anyone's interested, please take a look at the attached report and let me

PR #3, opened 2017-11-16
Ajc cell barcode correction

ambrosejcarr:
This is the tool portion of ticket 300 

It adds two new classes: 
`barcode.ErrorsToCorrectBarcodesMap`
`fastq.BarcodeGeneratorWithCorrectedCellBarcodes`

With these two classes, it was possible to add a `--whitelist` parameter to `Attach10XCellBarcodes`, the command-line function that attaches 10x cell barcodes. When the whitelist parameter is passed, `CB` tags (the htslib standard tag for corrected cell barcodes) are added to any record that has a barcode within 1 hamming distance from the whitelist. If the whitelist is not passed, the old functionality is retained (no correction). 

The new classes are fully tested, and you can see what is and is not tested by running: 
```
cd sctools
pytest --cov-report html:cov_html --cov=sctools
open cov_html/index.html
```
** This will open a browser window.


src/sctools/platform.py
```diff
@@ -6,45 +6,40 @@ class PlatformBase:
...
-    def get_tags(cls, sequencing_read):
+    def attach_barcodes(cls, args=None):
+        """command line entrypoint for attaching barcodes to a bamfile."""
         raise NotImplementedError
 
+
+class TenXV2(PlatformBase):
+
+    cell_barcode = fastq.EmbeddedBarcode(start=0, end=16, quality_tag='CY', sequence_tag='CR')
+    molecule_barcode = fastq.EmbeddedBarcode(start=16, end=26, quality_tag='UY', sequence_tag='UR')
```
samanehsan: Where do these numbers come from? Also what do these tag abbreviations `CY`, `UY`, etc. stand for?

ambrosejcarr: These are 10x v2 chemistry definitions. The tags are (soon to be) GA4GH standard tags for cell and molecular barcodes and their qualities. Would it be helpful to add links? Comments?

src/sctools/platform.py
```diff
@@ -6,45 +6,40 @@ class PlatformBase:
...
-    }
-
-    @classmethod
-    def get_tags(cls, sequencing_read):
-        return cls._tags[sequencing_read]
+    def make_tag_generators(cls, r1, i1, whitelist=None):
+        if whitelist is not None:
+            return [
+                fastq.BarcodeGeneratorWithCorrectedCellBarcodes(
+                    r1, cls.cell_barcode, whitelist, cls.molecule_barcode),
```
samanehsan: Can you pass these in with parameter names?

src/sctools/platform.py
```diff
@@ -6,45 +6,40 @@ class PlatformBase:
...
-
-    _tags = {
-        'r1': (cell_barcode, molecule_barcode),
-        'i1': (sample_barcode,)
-    }
-
-    @classmethod
-    def get_tags(cls, sequencing_read):
-        return cls._tags[sequencing_read]
+    def make_tag_generators(cls, r1, i1, whitelist=None):
```
samanehsan: Can you add a docstring describing what this method is doing and what the expected inputs/outputs are?

src/sctools/platform.py
samanehsan: I think comments would be helpful for clarifying what the tags are for the barcodes. For the start and end numbers, you could put those into variables that describe what they are, e.g. `CELL_BARCODE_END= 16` (or whatever name makes sense to you)

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+        """use a hash map of erroneous_barcode -> true barcode to correct errors in cell barcodes
+
+        :param dict errors_to_barcodes:
+        """
+        if not isinstance(errors_to_barcodes, Mapping):
+            raise TypeError('errors_to_barcodes must be a mapping of erroneous barcodes to correct '
+                            'barcodes, not %a' % type(errors_to_barcodes))
+        self._map = errors_to_barcodes
+
+    def correct_barcode(self, barcode):
```
samanehsan: Maybe rename this to `get_correct_barcode` to clarify that this just returns the correct barcode

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+        return self._map[barcode]
+
+    @staticmethod
+    def _prepare_single_base_error_hash_table(iterable):
```
samanehsan: I don't think the parameter name `iterable` is very descriptive. How about calling it `barcodes` instead?

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+
+        :param str bam_file:
+        :param str output_bam_file:
+        """
+        with pysam.AlignmentFile(bam_file, 'rb') as fin, \
+                pysam.AlignmentFile(output_bam_file, 'wb', template=fin) as fout:
+            for alignment in fin:
+                try:
+                    tag = self.correct_barcode(alignment.get_tag('CR'))
+                    alignment.set_tag('CB', tag, 'Z')
```
samanehsan: Can you pass these in with their parameter names? I'm not familiar with pysam so I'm not sure what `Z` is

samanehsan: Also it's clearer to only include code in the `try` block that you expect to throw the error, which looks like it's `tag = self.correct_barcode(alignment.get_tag('CR'))`

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+        :param str whitelist_file:
+        :return CorrectBarcodeErrorsHash:
+        """
+        with open(whitelist_file, 'r') as f:
+            return cls(cls._prepare_single_base_error_hash_table((line[:-1] for line in f)))
+
+    def correct_bam(self, bam_file, output_bam_file):
+        """correct barcodes in a (potentially unaligned) bamfile, given a whitelist
+
+        :param str bam_file:
```
samanehsan: Can you add descriptions for these parameters as well?

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+                errors.discard(nucleotide)
+                for e in errors:
+                    error_map[barcode[:i] + e + barcode[i + 1:]] = barcode
+        return error_map
+
+    @classmethod
+    def single_hamming_errors_from_whitelist(cls, whitelist_file):
+        """factory method to generate instance of class from a file containing "correct" barcodes
+
+        :param str whitelist_file:
```
samanehsan: Same as below, can you please add descriptions for these parameters?

src/sctools/fastq.py
```diff
@@ -160,41 +160,83 @@ def __iter__(self):
...
+        parse fastq files for barcodes(s) defined by EmbeddedBarcode objects, producing a generator
+        that yields extracted barcode objects in a form that is consumable by pysam's bam and sam
+        set_tag methods.
+
+        :param list|str fastq_files: (default sys.stdin) fastq file or list of files to be read.
+        :param EmbeddedBarcode embedded_cell_barcode: cell barcode tag
+        :param [EmbeddedBarcode] other_embedded_barcodes: list of tag objects defining start and
+          end of the sequence containing the tag, plus the string quality and sequence tags
+        :param str whitelist: whitelist file containing "correct" cell barcodes for an experiment
+        :param mode: (Default 'r') will return string objects. Change to 'rb' to
```
samanehsan: Is the `mode` parameter getting used anywhere?

src/sctools/fastq.py
```diff
@@ -160,41 +160,83 @@ def __iter__(self):
 EmbeddedBarcode = namedtuple('Tag', ['start', 'end', 'sequence_tag', 'quality_tag'])
 
 
+def extract_barcode(record, eb):
```
samanehsan: I think it would be clearer to avoid short abbreviations for variable names where possible. In this case I find `eb` confusing because I thought it stood for `extracted_barcode` but it's actually `embedded_barcode`

README.rst
```diff
@@ -0,0 +1,61 @@
+sctools
+=======
+
+sctools provides utilities for manipulating sequence data formats suitable for use in distributed
+systems analyzing large biological datasets.
+
+Download and Installation
+-------------------------
+
+.. code bash
```
samanehsan: Should this be uncommented or removed?

README.rst
```diff
@@ -0,0 +1,61 @@
...
+6. **fastq.Reader & fastq.Record** fastq reader and fastq record class that exposes the fastq fields
+   as a lightweight, lazy-parsed python object.
+
+
+Viewing Test Coverage
+---------------------
+to calculate and view test coverage cd to the ``sctools`` directory and
+type the following two commands to generate the report and open it in your web browser:
+
+.. code bash
```
samanehsan: Should this be uncommented or removed?

src/sctools/barcode.py
```diff
@@ -112,3 +113,66 @@ def from_iterable_bytes(cls, iterable, barcode_length):
...
+        :return CorrectBarcodeErrorsHash:
```
samanehsan: How is this different from `ErrorsToCorrectBarcodesMap`?

src/sctools/test/test_barcode.py
```diff
@@ -1,11 +1,15 @@
...
 import numpy as np
 import pytest
+import pysam
+from .. import barcode, encodings, platform
+
 
 data_dir = os.path.split(__file__)[0] + '/data/'
 
 
+# TEST BARCODES
```
samanehsan: Can you divide these tests into different test classes (e.g. `TestBarcodes`, `TestHashErrorsToCorrectBarcodes`, etc.)?

src/sctools/test/test_barcode.py
```diff
@@ -58,3 +62,81 @@ def test_summarize_hamming_distances_gives_reasonable_results(short_barcode_set_
...
     # no barcode can have more hamming distance than length
     assert all(v <= 16 for v in hamming_summary.values())
+
+
+# TEST HashErrorsToCorrectBarcodes
+
+@pytest.fixture(scope='module')
+def trivial_whitelist():
+    barcode_iterable = ['A' * 8]
+    em = barcode.ErrorsToCorrectBarcodesMap._prepare_single_base_error_hash_table(barcode_iterable)
```
samanehsan: I think `error_mapping` would be a clearer variable name here

src/sctools/test/test_barcode.py
```diff
@@ -58,3 +62,81 @@ def test_summarize_hamming_distances_gives_reasonable_results(short_barcode_set_
...
+    return barcode.ErrorsToCorrectBarcodesMap(em)
+
+
+@pytest.fixture(scope='module')
+def truncated_whitelist_from_10x():
+    # note that this whitelist contains 1 non-10x barcode to ensure the presence of a matching
+    # target in the test data.
+    ec = barcode.ErrorsToCorrectBarcodesMap.single_hamming_errors_from_whitelist(
```
samanehsan: Can you also replace `ec` with what it stands for? Sorry to nit-pick your variable names!

src/sctools/test/test_entrypoints.py
```diff
@@ -24,3 +24,29 @@ def test_Attach10XBarcodes_entrypoint():
...
+
+def test_Attach10XBarcodes_entrypoint_with_whitelist():
+    args = [
+        '--r1', data_dir + 'test_r1.fastq',
+        '--i1', data_dir + 'test_i7.fastq',
+        '--u2', data_dir + 'test_r2.bam',
+        '--output-bamfile', 'test_tagged_bam.bam',
+        '--whitelist', data_dir + '1k-august-2016.txt']
+
+    rc = platform.TenXV2.attach_barcodes(args)
```
samanehsan: Hopefully the last one here...but what is `rc`?

src/sctools/test/test_fastq.py
```diff
@@ -188,3 +188,57 @@ def test_fields_populate_properly(reader_all_compressions):
         assert all(i in alphabet for i in record.sequence.strip())
         assert record.name2 == name2_string
         assert all(i in ascii_chars for i in record.quality.strip())
+
+
+# TEST BarcodeGeneratorWithCorrectedCellbarcodes
```
samanehsan: Can you group these tests into a test class `TestBarcodeGeneratorWithCorrectedCellbarcodes`?

src/sctools/test/test_fastq.py
```diff
@@ -188,3 +188,57 @@ def test_fields_populate_properly(reader_all_compressions):
...
+        data_dir + 'test_r1.fastq.gz', cell_barcode, data_dir + '1k-august-2016.txt',
+        molecule_barcode)
+
+
+def test_embedded_barcode_generator_produces_outputs_of_expected_size(embedded_barcode_generator):
+    for cell_seq, cell_qual, umi_seq, umi_qual in embedded_barcode_generator:
+
+        # test cell tags
+        assert cell_seq[0] == 'CR'
+        assert len(cell_seq[1]) == 16
```
samanehsan: Can you replace these hard-coded numbers with variables? You can make them properties of the test class

src/sctools/test/test_fastq.py
```diff
@@ -188,3 +188,57 @@ def test_fields_populate_properly(reader_all_compressions):
...
+        break  # just the first tag is fine
+
+
+def test_corrects_barcodes(barcode_generator_with_corrected_cell_barcodes):
+    success = False
+    for barcode_sets in barcode_generator_with_corrected_cell_barcodes:
+        for barcode_set in barcode_sets:
+            if barcode_set[0] == 'CB':
+                success = True
+    assert success
```
samanehsan: This will pass if at least one of the barcodes is corrected, but not necessarily all of them. Is that intended?

samanehsan:
Looks good @ambrosejcarr! I mostly have small formatting/naming suggestions :octocat:

README.rst
ambrosejcarr: That's how ReST-based markdown works -- it's not a comment; it's boilerplate to initiate a code block. We unfortunately need ReST for uploads to Pypi. :(

src/sctools/test/test_entrypoints.py
ambrosejcarr: Return Code. I'll be explicit.

src/sctools/test/test_fastq.py
ambrosejcarr: Yep; it's intended. There is actually only one barcode capable of correction in the test data >.>

samanehsan: Oh ok! In that case you can `break` after you find the corrected barcode

src/sctools/test/test_entrypoints.py
samanehsan: thanks!

README.rst
samanehsan: Ohhh okay! Sorry, I just assumed it was a comment because the text is grey :/

ambrosejcarr: No need to apologize :)

src/sctools/platform.py
ambrosejcarr: I think the fact that it's an embedded barcode object with a start and end is maybe enough here, instead of explicitly expanding out into another 12 variables. 

That said, I've added a comment above to describe what they're doing -- if that's not enough to address your concern I'll reconsider expanding the variables. Let me know!

src/sctools/barcode.py
ambrosejcarr: the parameters are tag, tag_value, tag_type -- is that helpful? My temptation is to add a comment which might be more informative.

ambrosejcarr: Also, I normally agree, but in this case I'm only setting the tag if it is discovered; if I move the set call outside, I need another `if` boolean, and so here I left it inside to improve efficiency.

src/sctools/fastq.py
ambrosejcarr: Yep! I had forgotten to include *args and **kwargs in init; it gets passed to the super class (reader) if it's present.

src/sctools/fastq.py
ambrosejcarr: Did it to avoid multi-lining some of the commands; turns out it was fine with the full name, too. :)

src/sctools/barcode.py
ambrosejcarr: Good catch; failed to refactor the docstrings during development. It should be as you say.

src/sctools/test/test_barcode.py
ambrosejcarr: I think the answer is "yes, but it'll be complicated": http://computableverse.com/blog/pytest-sharing-class-fixtures. I'm hesitant to do the refactor in this ticket.

src/sctools/test/test_fastq.py
ambrosejcarr: Added some comments, but generally I want to avoid unnecessary verbosity by duplicating all of these.

src/sctools/test/test_fastq.py
ambrosejcarr: Same answer as above.

ambrosejcarr:
I think I got all the comments. Back to you @samanehsan

src/sctools/platform.py
samanehsan: I think this looks good!

src/sctools/barcode.py
samanehsan: Oh, I see you `pass` if there's a KeyError so if you want to set the tag after it's not necessarily there

samanehsan: I think parameter names would make it so that you don't have to add a comment about what each one is, but you could certainly do both

src/sctools/platform.py
```diff
@@ -6,45 +6,64 @@ class PlatformBase:
...
+        """Create tag generators from fastq files.
+
+        Tag generators are iterators that run over fastq records, they extract and yield all of the
+        barcodes embedded in each fastq record. For 10x, this means extracting the cell, umi, and
+        sample barcode.
+
+        :param str r1:
```
samanehsan: Can you include a description for these params?

src/sctools/test/test_barcode.py
samanehsan: Ok, I'm not familiar with pytest but this would be a good code-cleanup task since grouping tests into test classes helps keep the code organized, especially as you add in more tests!

ambrosejcarr: Definitely agree, but I ran into some pretty significant weirdness when trying it with fixtures. There weren't good examples of how to use fixtures inside classes (since fixtures get passed as parameters, but become methods of cls...) Once that gets sorted out, we can probably fix it. However, it's a 2-point ticket involving a bit of research on how other projects use pytest (and if they use classes, or maybe have their own files, per class that is being tested). 

I've made a ticket.

src/sctools/fastq.py
```diff
@@ -160,41 +160,88 @@ def __iter__(self):
...
-    def extract_barcode(record, eb):
-        """extracts barcodes from a fastq record according to an EmbeddedBarcode object.
-
-        :param FastqRecord record: record to extract from
-        :param EmbeddedBarcode eb: defines the barcode start and end and the names of the quality
-          and sequence tags
-        :return tuple: (sequence tag name, tag content (nucleotides), tag type (Z))
-        :return tuple: (quality tag name, tag content (qualities), tag type (Z))
+
+class BarcodeGeneratorWithCorrectedCellBarcodes(Reader):
```
samanehsan: Why have a separate class for handing corrected cell barcodes instead of modifying `EmbeddedBarcodeGenerator`?

src/sctools/test/test_barcode.py
samanehsan: Ok! Thanks for making a ticket!

src/sctools/fastq.py
ambrosejcarr: So, I initially subclassed `EmbeddedBarcodeGenerator` instead of `Reader`, but that did weird stuff to `__init__`, which I would also have to overload. Since I was basically overloading all the methods, it made more sense to do a parallel subclassing of reader. 

I created a separate method because it didn't seem worth it to generalize to `CorrectedEmbeddedBarcodeGenerator` at the time; I would need to pair whitelists with barcodes in a separate class, then find a way to bring in barcodes with and without construction. This should be possible, but I feel I need to focus on creating a draft pipeline for the quarterly demo. I'm happy to make a tech debt ticket to combine these classes If I get all the pre-demo work done with time to spare. (?)

What do you think?

samanehsan: Ok, thanks for the explanation! I think adding a follow-up ticket is fine

ambrosejcarr: Ok. In that case, I will leave the follow-up branch open -- there is potential for some debugging to be needed when I apply to it to wdl, and because of how Pypi works, I need to  increment the version each time I push it up. 

We can revisit merging the PR when all of the WDL coding is completed.

ambrosejcarr: Ticket made.

PR #5, opened 2017-12-07
Ajc fix compression detection

ambrosejcarr:
- fixed sctools to detect compressed files based on magic numbers, not rely on file extensions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ CLASSIFIERS = [
 
 setup(
     name='sctools',
-    version='0.1.4',
+    version='0.1.6',
     description='Utilities for large-scale distributed single cell data processing',
     url='https://github.com/humancellatlas/sctools.git',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -17,7 +17,7 @@
 
 setup(
     name='sctools',
-    version='0.1.4',
+    version='0.1.6',
```
samanehsan: Why is this version `0.1.6` and not `0.1.5`?

ambrosejcarr: Because I need to start using the pypi test server instead of pushing things with bugs directly to pypi's main index. :)

samanehsan: Ohh okay! What was the bug? Do you think additional tests could help or was it really minor like a typo?

ambrosejcarr: Yea.. I think there was a testing failure. I'll think about how to add them. 👍

PR #6, opened 2018-02-06
Base work for Optimus Metrics

jsotobroad:
@ambrosejcarr after we get comments on this tidied its up to you if we should merge it in or just use the branch as the base for our future work.  I'm ok either way.


src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+        for metric in metrics_to_run:
+            metric.setup()
+
+        with pysam.AlignmentFile(self.input_sam, self.open_mode) as sam:
+            for record in sam:
+                for metric in metrics_to_run:
+                    metric.gather_metric(record)
+
+        for metric in metrics_to_run:
+            metric.calculate_and_output()
```
ambrosejcarr: I think we might want to split up `calculate` and `output`. I think the pattern as we've discussed is to have each metric class broken up by what its inputs are, but the _outputs_ might be split differently. I could imagine everything that produces a vector summarizing all the genes to want to go into the same output matrix, for example. 

Maybe some kind of general `output` that takes a bunch of calculated metrics in a particular shape and outputs them together? 

Thinking about memory usage, if those `calculate` commands generate large data structures that need to be dumped immediately to disk, then `output` could open up a file object and take a bunch of metrics as parameters, which it would then `calculate`, dump to the file, and grab the next one. 

... or is this overthinking because we could just concatenate new rows to a .csv.gz file? In any case, I want some kind of thoughtful solution here. :)

src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def convert_class_name_to_class(metric_class_names):
```
ambrosejcarr: Good idea. I think this is just `getattr(module, name)` or `operator.attrgetter` if we want it to have a custom name. 

https://docs.python.org/3.7/library/operator.html#operator.attrgetter

src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+    # maybe some kind of case statement or python equivalent to grab a class from a string
+    # once we define a metric class that extends basemetric.py we can fill out this method
+    pass
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("input_bam", help="path to input bam file")
+    parser.add_argument("-m", "--metric", action="append", help="metrics to run on input_bam")
```
ambrosejcarr: I'm a fan of `-m [list of whitespace-separated metrics]` vs `-m [m_1] -m [m_2] ... -m [m_n]`. What're your thoughts? 

In the former case, I'd make it `metrics` so that it's more obvious when referencing that we want multiple (see below)

src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+    args = parser.parse_args()
+
+    metric_classes = None  # convert_class_name_to_class(args.metric)
```
ambrosejcarr: I think a neat way of doing this would be as follows (with better names for the directory structure): 

Make the structure something like this: 
```
metrics
__init__.py
runner.py
base_metric.py
classes/__init__.py
classes/metric_one.py
classes/metric_two.py
...
classes/metric_n.py
```

**`metrics/classes/__init__.py:`**
```
from .metric_one import metric_one
from .metric_two import metric_two
...
from .metric_n import metric_n
```
**`metrics/runner.py:`**
```
from metrics import classes
...

metrics = [getattr(classes, metric_name) for metric_name in metrics]
```

src/sctools/metrics/metricsrunner.py
ambrosejcarr: I think we can just use `getattr` here (or operator.attrgetter if we want a custom name). See below, I have an idea about how to implement and structure this.

src/sctools/metrics/README.md
```diff
@@ -0,0 +1,44 @@
...
+#### BaseMetric - base class for different metrics collectors
+* setup() - any setup a metrics class has to do
+* accept_read() - process a record for your metric
+* calculate_and_output() - finalize any calculations needed and write out file
+
+
+#### MetricsRunner - class with command line interface, determines what metrics to run and inputs for those metrics
+* loop through each metric and call setup() on each
+* loop through bam calling metric.accept_read() for each metric
+* loop through each metric and call finish_up() on each
```
ambrosejcarr: As discussed, :puke: :)

src/sctools/metrics/basemetric.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class BaseMetric:
+
+    @abc.abstractmethod
+    def initialize(self):
+        """Do some kind of one time initialization for the metrics class"""
+
+    @abc.abstractmethod
+    def gather_metric(self, sam_record):
```
ambrosejcarr: This should get a ReST docstring as well.

src/sctools/metrics/basemetric.py
```diff
@@ -0,0 +1,16 @@
...
+        """What the metric to do with each read of a bam"""
+
+    @abc.abstractmethod
+    def calculate_and_output(self):
```
ambrosejcarr: This probably needs some parameters (like a file handle) -- that said I have some comments about this below to read first.

src/sctools/metrics/README.md
jsotobroad: I could have sworn I changed this!

src/sctools/metrics/basemetric.py
```diff
@@ -0,0 +1,16 @@
+import abc
```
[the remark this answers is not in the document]
jsotobroad: will change to this pattern, I dont have a real preference and will take all the tips i can get

src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+class MetricsRunner:
+
+    def __init__(self, sam_file, open_mode=None):
+        """
+        :param str alignment_file: sam or bam file.
+        :param str open_mode: optional, mode to read file. Will be autodetected by file type if
+          the file contains the correct suffix for its type.
+        """
+
+        if open_mode is None:
```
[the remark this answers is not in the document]
jsotobroad: Hmm I believe a bam is block compression on top of standard gzip so we **should** be able to do something fancy here?  I'm not too worried about this though at the moment.

src/sctools/metrics/metricsrunner.py
```diff
@@ -0,0 +1,61 @@
...
+        self.input_sam = sam_file
+        self.open_mode = open_mode
+
+    def run_metrics(self, metrics_to_run):
+        """
+        :param list metrics_to_run: list of metrics to run for each record in input_sam
+        """
+
+        for metric in metrics_to_run:
+            metric.setup()
```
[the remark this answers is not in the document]
jsotobroad: 💯  totally forgot to follow through with the refactor.

src/sctools/metrics/metricsrunner.py
jsotobroad: The idea I had was that each metric would be given some kind of basename metric name and then each metric would output however many files it needs to given the basename.  I was thinking `.tsv` or `.csv` should be good memory friendly enough.

I think this should still just be one method because at this point our memory footprint would be at its largest and we just want to go through each metric and just calculate whatever we need to and have it write out the files.  Once its written its files all of its members could then be cleaned up as we move on and do the same to the next metric.  I wouldn't want to `calculate` on every metric making out memory footprint larger than it has to be.  I definitely think we should think this through though.

src/sctools/metrics/metricsrunner.py
jsotobroad: So the issue I had with`-m [list of whitespace-separated metrics]` is that it seemed you had to know how many inputs you were expecting but i took a second look at the docs and it seems its really easy.  I think I prefer being able to dump all the metrics at once for now so will change to `metrics`

src/sctools/metrics/metricsrunner.py
jsotobroad: definitely was not thinking about structure yet but I'm not sure if `/classes/` is the best sub dir name but I cant think of a better one so we'll stick with it for now!  I'm not even sure if `metrics` should be the root dir name like it currently is.  Always room for improvement.

You will probably have to explain to me how `from metrics import classes` works 😬 .

jsotobroad:
I hope to get a lot more things flushed out as we start writing metrics!

jsotobroad:
@ambrosejcarr back to you

src/sctools/metrics/metricsrunner.py
ambrosejcarr: It triggers the `classes/__init__.py` and dumps the results into the `classes` namespace, now in your global namespace.

ambrosejcarr: so `classes.metric_one` becomes addressable, which means that getattr(classes, "metric_one") will produce `metric_one`.

src/sctools/metrics/metricsrunner.py
ambrosejcarr: Not 100% sure what you mean by `basename` here. We definitely will want any comparable metrics to sit in the same file so they can be accessed via PCA or something similar. 

That said, I think the latter part of my ramblings above identifies that we can just open up a file in append mode and dump another row into it, which accomplishes both our requirements.

src/sctools/metrics/metricsrunner.py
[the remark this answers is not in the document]
ambrosejcarr: Cool; I only mention it because every time I haven't done this it breaks the pipeline due to WDL/Cromwell not playing nice with file handles.

PR #8, opened 2018-02-07
made index read optional

ambrosejcarr:
This is the cognate to the Optimus PR. This allows Attach10xBarcodes to function without receiving i1 fastq file(s).

### DIFF
--- a/src/sctools/platform.py
+++ b/src/sctools/platform.py
@@ -69,7 +69,7 @@ class TenXV2(GenericPlatform):
         bam_tagger.tag(output_bamfile_name, tag_generators)
 
     @classmethod
-    def make_tag_generators(cls, r1, i1, whitelist=None):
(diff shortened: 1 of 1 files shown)


jishuxu:
Good looks. I have a question. If I1 is not required, do you need to generate random mocked up barcodes to replace I1? Or Do you need I1 at all?

ambrosejcarr:
@jishuxu i1 is not currently required by any downstream steps of the pipeline. When we eventually allow for data matrices to be split by sample index, I believe a simple check for whether or not the i1 tag is present in the bam will be enough to trigger the correct logic. 

There is no need for any i1 sample tags to my knowledge. If we discover a need later, I can use the `tag_generator` class to produce a generator that just produces `NNNNNNNN` for every r1 and u2 read.

PR #10, opened 2018-02-12
First pass at unique fragments per umi metric

jsotobroad:


src/sctools/metrics/classes/collect_unique_fragment_by_umi.py
```diff
@@ -0,0 +1,74 @@
...
+
+class UniqueFragmentPerUMI(base_metric.BaseMetric):
+
+    def __init__(self):
+        self.cell_barcode_to_umi = None
+        self.basename = None
+        self.cell_barcode_tag = None
+        self.umi_barcode_tag = None
+
+    def initialize(self, args):
```
ambrosejcarr: I think this can be rolled into `__init__`

src/sctools/metrics/classes/collect_unique_fragment_by_umi.py
```diff
@@ -0,0 +1,74 @@
...
+        self.cell_barcode_to_umi = collections.defaultdict(dict)
+        self.basename = args.basename
+        self.cell_barcode_tag = args.cell_barcode_tag
+        self.umi_barcode_tag = args.molecular_barcode_tag
+
+    def gather_metric(self, sam_record):
```
ambrosejcarr: I wonder if we should make this a holistic function that builds a data model that we can easily interact with _across_ metrics. I.e. if we should move it into the base class. 

This could be implemented a bunch of different ways -- there could be a flexibility build into the "fragment" class that I propose in my PR where we construct _that_ object on the fly and fill it according to what the user says they want us to record. 

However, I suspect we're going to want some kind of model wherein we can query information on fragments. 

Something like: 

`data_model = {(cell, umi, gene, (chr, pos, strand)): fragment_data}`

where fragment_data has some flexibility. From the gene perspective I know I want to track: 

fragment_data <- (n_observations, min_mismatches, is_unique, is_spliced)

We may also want some kind of average or maximum quality score for the fragment (fragments that are never covered by a high-quality read could be filtered out, for example)

some other things not covered yet by this model: 
- fragment entropy (could be calculated, is a bit computationally heavy)
- explicit model for how to deal with multi-aligned reads
- distance from 3' end (requires gtf parsing) **
- augmentation with metrics that require PCA, diffusion analysis, higher-order distance and covariance analyses (probably run on count matrix)


** look up if there is a good parser somewhere

jsotobroad: Interesting, so the obvious thing to be afraid of here is generating a huge data model where only a few of the metrics actually use all of the data.  Like for instance this metric we wouldn't need `gene, chr, pos, strand` information.  We could split our metrics up based on which share these models and generate that once per set of sharing metrics (I think that is what you are getting at, I just like repeating things :P).

ambrosejcarr: > We could split our metrics up based on which share these models and generate that once per set of sharing metrics

Yes exactly. I think we'll get into a situation where almost everything needs the base model for aggregating `fragment` information, but the individual pieces of fragment information may only be required by certain metrics.

PR #11, opened 2018-02-23
Ajc metric iterators

ambrosejcarr:
# Description of implementation for Metrics. Readme is copied below, for simplicity. 

Links to https://github.com/HumanCellAtlas/skylab/tree/ajc-gene-metrics-552 which contains WDLs to execute these functions.

## Overall Metric Processing
Metric processing, as proposed, relies on the structure of the Optimus pipeline to produce a metric accumulator that will scale with the size of an individual cell (a small, relatively statically sized quantity in 10x). It accomplishes this by sorting the files so that cell information is contiguous and running iterators that process cells and then dump the metrics to disk before proceeding to the next cell. Specifically, it:

1. relies upon chunking the input bam files into smaller pieces
2. relies on the guarantee of the optimus pipeline that for each cell, all reads associated with it
are found in the same chunk
3. Relies on sorting the (small) bam files based on different tags by first converting them to sam. 
The distributed nature of the pipeline makes this practical.
4. This sorting enables us to process all of the data for a molecule at one time. This is important, 
as many metrics integrate over a molecule's data. Without sorting, a data structure must be built to 
build up the information from the bam file, and the resulting memory footprint is 20-40gb for a 
standard HiSeq run.  
5. To sort, the file needs three tags: Cell barcode (CB), Molecule barcode (UB) and Gene (GE). The 
latter tag is not, to my knowledge, in the HTSlib spec. 

Despite the reliance on Optimus assumptions, We will be able to release this tool as a stand alone 
that can run on an arbitrary, properly tagged bam file, but it will rely on other sctools 
functionality. Specifically, it requires a set of tools that implement a map-reduce paradigm. For metrics, this means: 

1. Partition original bam into groups of cells
2. Sort partitioned files
3. Process sorted partitions
4. Merge results

An abbreviated WDL could be made in the future which would execute just the metrics portion of Optimus, and would contain: 

```
1. SplitBamByCellBarcode
2. scatter[CalculateMetrics]
3. MergeMetrics
```
It could also be possible to run this implementation on other frameworks by using the sctools docker, which is not tied to WDL/cromwell. 

## Implementation: 

I believe the best course of action is to generate metrics files that align with our eventual 
count matrix. This is consistent with scanpy/loom formats which use DataFrame-like objects to store
metadata for both samples and features. It will enable the most common access patterns of comparing metrics across the relevant data, and more complex access patterns like the fitting of mixed-effect models to correct for metadata values. 

This means we need two metrics outputs: one for Cells, and one for genes. 

I've pseudo-coded three classes, each of which must be subclassed to write a Metric: 
```
MetricAggregator:
  - CellMetricAggregator
  - GeneMetricAggregator

MetricGatherer:
  - CellMetricGatherer
  - GeneMetricGatherer

MergeMetrics:
  - MergeCellMetrics
  - MergeGeneMetrics
```

It also relies upon the `SplitBamByCellBarcode` tool, which was previously implemented in `sctools` to enable partitioning of the input data. 

MetricAggregator stores the information for a unit of the relevant data (cell, gene) as records are
being processed, and has csv writing capability so that the information for only one unit is stored
at a time, ensuring scalability to extremely large datasets. It has a `parse_records()` function 
which has not yet been implemented which will do most of the work

MetricGatherer defines iterables to do the record grouping, the bam parsing pattern necessary, 
and a `check_sort_order` function to verify the file is in the required order. 

MergeMetrics merges multiple metric outputs from the scattered chunks. This is a trivial 
concatenation in the case of cell metrics, and a more complex weighted averaging merge in the case of gene metrics, since gene information is split across the partitioned files. 

For anyone interested, there is a significant amount of code present in this PR that produces a working implementation for Gene Metrics (subject to verification of output correctness).


src/sctools/metrics/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```
+1. SplitBamByCellBarcode
+2. scatter[CalculateMetrics]
+3. MergeMetrics
+```
+
+## Implementation: 
+
+I believe the best course of action is to generate a metrics file that aligns with our eventual
```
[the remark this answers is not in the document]
ambrosejcarr: Fair. Documentation could use a rework, I wrote that as a proposal and forgot to go back. Thanks for the reminder.

src/sctools/metrics/merge.py
```diff
@@ -0,0 +1,51 @@
...
+        concatenated_frame: pd.DataFrame = pd.concat(metric_dataframes, axis=0)
+        concatenated_frame.to_csv(self._output_file)
+
+
+class MergeGeneMetrics(MergeMetrics):
+
+    # todo this class needs to do a weighted mean; this is something numpy can do.
+    def execute(self) -> None:
+        """
+        Currently this class returns an incorrect result
```
samanehsan: Do we want to use this if it's incorrect?

samanehsan: As discussed with @ambrosejcarr, this part is fixed in a different branch

src/sctools/metrics/aggregator.py
```diff
@@ -0,0 +1,319 @@
...
+        self.genomic_read_quality_mean: float = self._genomic_read_quality.mean
+
+        self.genomic_read_quality_variance: float = self._genomic_read_quality.calculate_variance()
+
+        self.n_molecules: int = len(self._molecule_histogram.keys())
+
+        self.n_fragments: int = len(self._fragment_histogram.keys())
+
+        try:
+            self.reads_per_molecule: float = self.n_reads / self.n_molecules
```
samanehsan: Just wondering, when would you have zero molecules?

src/sctools/test/test_metrics.py
```diff
@@ -0,0 +1,374 @@
...
+
+# todo fails, not yet properly set up
+def test_merge_gene_metrics():
+    """test the sctools merge gene metrics CLI invocation"""
+    TenXV2.merge_gene_metrics(
+        args=['-o', _test_dir + '/merged-gene-metrics.csv',
+              'gene_metrics_1.csv', 'gene_metrics_2.csv'])
+
+
+@pytest.mark.parametrize('metrics, expected_value', [
```
samanehsan: This is really cool!

src/sctools/test/test_metrics.py
```diff
@@ -0,0 +1,374 @@
...
+    a fragment covered by a single read
+    """
+    assert np.all(metrics['n_molecules'] >= 1)
+    assert np.all(metrics['n_fragments'] >= 1)
+    assert np.all(metrics['n_fragments'] >= metrics['n_molecules'])
+
+
+@pytest.mark.parametrize('metrics, key, expected_value', [
+    (_cell_metrics, 'molecule_barcode_fraction_bases_above_30_mean',
+     np.array([1.0000, 0.9500, 1.0000, 1.0000, 0.9778, 1.0000, 1.0000, 1.0000, 0.9833, 1.0000,
```
samanehsan: Are these just random inputs you created for test purposes?

ambrosejcarr: No, they are not. We examined the test data in the notebooks enclosed in this PR. We used different tools to arrive at what we expect the metric outputs to be and verify that the metric iterators defined for the production pipelines match our expectations.

samanehsan: Ok, thanks for the clarification! Can you add this in as a comment to clarify where this test data is coming from?

src/sctools/metrics/aggregator.py
ambrosejcarr: When every read for a cell is unmapped.

ambrosejcarr:
@samanehsan I've added documentation that address your questions if you'd like to take a look.

src/sctools/test/test_metrics.py
```diff
@@ -9,6 +9,14 @@
 from sctools.metrics.gatherer import GatherGeneMetrics, GatherCellMetrics
 from sctools.platform import TenXV2
 
+# ############################### Testing Ground Truth Values ################################### #
```
samanehsan: Minor style suggestion: you could use a `"""triple-quoted string here for this comment block"""`

src/sctools/metrics/README.md
```diff
@@ -1,41 +1,35 @@
...
+architecture execution at production scale. Specifically, the workflow is as follows: 
+
+1. Chunk the input bam file using `SplitBam`, which generates several chunks, each of which is
+guaranteed to contain all data for any cell it contains
+2. Sort each chunk by cell, gene, and molecule tags to ensure that all the reads associated with 
+a molecule are stored sequentially by cell (`CalculateCellMetrics`) or by gene 
+(`CalculateGeneMetrics`)
+3. For each cell or gene, parse the information by molecule, which typically loads fewer than 
+10,000 records into memory at a time
+4. Call
```
samanehsan: Did you mean to add more to this step?

PR #12, opened 2018-03-01
Ajc cell metric iterators

ambrosejcarr:
* Updates bam iterators to treat "no tag" as a group and yield it (for reads that have no genes). Alternatively, provides a parameter to filter out those groups. 

* provides flow control for unmapped reads, and processing for them in CellMetrics


src/sctools/bam.py
```diff
@@ -213,33 +213,37 @@ def _cleanup(files_to_counts, files_to_names, rm_all=False):
...
-            break
-        except KeyError:  # todo this needs to be tracked so we know how many reads were invalid
-            continue
-        except StopIteration:
-            raise ValueError('No reads contained tag %s' % tag)
+    read = next(bam_iterator)
+    try:
+        current_tag = read.get_tag(tag)
+    except KeyError:
+        current_tag = None  # null tag is a category that gets emitted
```
jsotobroad: I think we should be checking for `filter_null` here in these `except` blocks and either `continue`ing in the bottom case or throwing an error here if `true`.  This way we dont even have to the least amount of work if we're filtering them out anyways.

src/sctools/bam.py
```diff
@@ -213,33 +213,37 @@ def _cleanup(files_to_counts, files_to_names, rm_all=False):
...
             next_tag = alignment.get_tag(tag)
         except KeyError:
-            continue
+            next_tag = None  # null tag is a category that we will emit
         if next_tag == current_tag:
             reads.append(alignment)
         else:
-            yield iter(reads), current_tag
+            # only yield if the tag is non-null or filter_null is false
+            if not filter_null or current_tag is not None:
```
jsotobroad: if you do what the comment above suggests you can get rid of the if checks here.

ambrosejcarr: I prefer them here because the else clause will execute far less often than the if clause. Thus, faster code.

src/sctools/metrics/aggregator.py
```diff
@@ -134,13 +135,15 @@ def parse_molecule(
...
 
-            # call the parse_extra_fields function to extract subclass-specific information
-            # todo might need a callback if this function has knowledge about termination conditions
+            # extract sub-class-specific information
             self.parse_extra_fields(tags=tags, record=record)
 
             self.n_reads += 1
             # self.noise_reads += self.is_noise(record)
 
+            # need a distinction between aligned reads and unaligned reads
```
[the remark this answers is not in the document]
ambrosejcarr: None. I can normalize the language to only use "mapped"

src/sctools/metrics/aggregator.py
```diff
@@ -161,7 +164,11 @@ def parse_molecule(
                 self.quality_above_threshold(30, record.query_alignment_qualities)
             )
 
-            # alignment location information
+            # the remaining portions deal with mapped reads, so if the read is not mapped, drop it.
+            # todo verify that this does not break tests
```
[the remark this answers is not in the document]
ambrosejcarr: Correct, we'll need to use unmapped reads.

src/sctools/metrics/aggregator.py
```diff
@@ -251,13 +259,17 @@ def parse_extra_fields(self, tags: Sequence[str], record: pysam.AlignedSegment)
...
-        alignment_location = record.get_tag('XF')
-        if alignment_location == 'INTERGENIC':
-            self.reads_mapped_intergenic += 1
+
+        try:
+            alignment_location = record.get_tag('XF')
+            if alignment_location == 'INTERGENIC':
+                self.reads_mapped_intergenic += 1
+        except KeyError:
+            self.reads_unmapped += 1
```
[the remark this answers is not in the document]
ambrosejcarr: I don't think that's true; this gets called at the top of the `parse_molecule()` block before the `.is_unmapped` check is done, no?

src/sctools/bam.py
ambrosejcarr: See below for response.

PR #13, opened 2018-03-02
Debugging Gene Metrics

jsotobroad:


ambrosejcarr:
This is an alternative bam parser that (1) allows missing tags and (2) lets you read in the soft-clipped qualities. It has the disadvantage of being closer to the production code. 

```
def parse_record(record):
    """line parser to build dataframe, supports missing tags in test data"""
    data = {
        'qname': record.query_name,
        'flag': record.flag,
        'reference': record.reference_id,
        'position': record.pos,
        'mapq': record.query_qualities,
        'cigar': record.cigarstring,
        'rnext': record.rnext, 
        'pnext': record.pnext,
        'tlen': record.tlen, 
        'sequence': record.seq,
        'quality': record.qual,
    }
    for name, tag in record.get_tags():
        data[name] = tag
    return pd.Series(data)

with pysam.AlignmentFile('data/cell-sorted.bam', 'rb') as f:
    records = []
    for record in f:
        records.append(parse_record(record))

data = pd.concat(records, axis=1).T
```

ambrosejcarr:
for 
```
results['reads_per_molecule'] = reads_per_molecule.mean()
results['reads_per_fragment'] = reads_per_fragment.mean()
results['fragments_per_molecule'] = fragments_per_molecule.mean()
```

Can we make these tests without the mean? It would be more stringent to explicitly check all the genes with `math.isclose()`

src/sctools/test/test_metrics.py
```diff
@@ -136,24 +138,33 @@ def test_spliced_reads(gene_metrics):
...
+        ('fragments_per_molecule', float)
     ]
     for (key, func) in higher_order_metrics:
         observed = gene_metrics[key].mean()
         expected = func(_testing_knowledge[key])
-        assert observed == expected, '%s failed' % key
+        assert math.isclose(observed, expected), '%s failed' % key
 
+
+def test_higher_order_metrics_sum(gene_metrics):
```
ambrosejcarr: Lets test equivalence for each of these and output a longer error message that includes all the failing tests at once. 

Something like: 
```
test_results = {}
for (key, func) in higher_order_metrics:
    observed = gene_metrics[key].sum()
    expected = func(_testing_knowledge[key])
    test_results[key] = (observed == expected, observed, expected)
# lazy but fine: print dictionary of results if any test fails. 
assert all(v[0] for v in test_results.values()), repr(test_results)  
```

src/sctools/metrics/gatherer.py
```diff
@@ -32,16 +32,16 @@ def extract_metrics(self, mode: str='rb') -> None:
             # write the header
             cell_metrics_output.write_header(vars(CellBarcodeMetrics()))
 
-            # break up the bam file into sub-iterators over cells
+            # break up the bam file into sub-iterators over genes
```
ambrosejcarr: These three changes here are wrong, I think... 

It should indeed be cell -> molecule -> gene

src/sctools/test/data/small-gene-sorted_testing_knowledge.csv
```diff
@@ -1,25 +1,95 @@
 duplicate_reads,51
-duplicate_records,51
-fragments_per_molecule,2.465909090909091
+fragments_per_molecule,"GE
```
ambrosejcarr: I feel like we probably don't want to commit this file.

src/sctools/test/test_metrics.py
```diff
@@ -52,78 +55,79 @@ def gene_metrics():
...
 
 
 def test_duplicate_records(gene_metrics):
     observed = gene_metrics['duplicate_reads'].sum()
-    expected = int(_testing_knowledge['duplicate_reads'])
+    expected = int(_testing_knowledge_scalar['duplicate_reads'])
     assert observed == expected
 
 
+    # todo this test currently fails because we believe our data is not deduped correctly, need to run on data run on latest optimus pipeline
```
ambrosejcarr: Lets change this comment to reflect that we're perplexed, not a failure of deduping.

src/sctools/metrics/gatherer.py
jsotobroad: oh did i change cell metrics gatherer instead of gene -_-.  thanks making the change

PR #16, opened 2018-03-16
Add merge functionality

ambrosejcarr:
* Fixes the broken `MergeGeneMetrics` command
* Adds the `merge` functions to the CLI
* Adds merge tests.

### DIFF
--- a/src/sctools/metrics/merge.py
+++ b/src/sctools/metrics/merge.py
@@ -71,12 +71,12 @@ class MergeGeneMetrics(MergeMetrics):
 
         sum_operations = {c: 'sum' for c in columns_to_sum}
 
-        def weighted_average(x):
+        def weighted_average(data_frame):
             """
-            :param pd.DataFrame x: input dataframe to reduce with a weighted average
+            :param pd.DataFrame data_frame: input dataframe to reduce with a weighted average
             :return pd.Series: reduced result
             """
-            weights = x['n_reads'].values
+            weights = data_frame['n_reads'].values
 
             columns_to_average_by_read = [
                 'molecule_barcode_fraction_bases_above_30_mean',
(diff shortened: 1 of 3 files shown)


src/sctools/metrics/merge.py
```diff
@@ -38,14 +40,87 @@ class MergeGeneMetrics(MergeMetrics):
...
+            'number_cells_detected_multiple',
+            'number_cells_expressing',
         ]
-        index: pd.Index = metric_dataframes[0].index
-        columns: pd.Index = metric_dataframes[0].columns
-        averaged_data: np.ndarray = np.mean(metric_dataframes, axis=0)
+
+        sum_operations = {c: 'sum' for c in columns_to_sum}
+
+        def weighted_average(x):
```
samanehsan: How about naming this `data_frame` instead of `x` for improved readability?

src/sctools/test/test_metrics.py
```diff
@@ -370,5 +356,92 @@ def test_single_read_evidence(metrics, key, expected_value):
...
+
+
+@pytest.fixture
+def mergeable_gene_metrics():
+    return split_metrics_file(_gene_metric_output_file)
+
+
+def test_merge_cell_metrics_cli(mergeable_cell_metrics):
+    """test the sctools merge cell metrics CLI invocation"""
+    TenXV2.merge_cell_metrics(
```
samanehsan: I'm not sure I understand this test -- should it assert something?

samanehsan:
I just have a few comments for you @ambrosejcarr otherwise this looks good!

src/sctools/test/test_metrics.py
ambrosejcarr: That's a good point. I had written the CLI tests to simply verify that they run. However, I've coded them all to return 0, so I'll verify that the correct return call is being generated. 👍

src/sctools/metrics/merge.py
ambrosejcarr: Done.

ambrosejcarr:
@samanehsan Comments have been addressed in the most recent commit if you want to take a look. Thanks for the input!

Issue #17, opened 2018-03-16
Cell barcode correction is wrong


ambrosejcarr:
Right now, we do this to identify a cell barcode: 
```
    def extract_cell_barcode(self, record, cb):
        seq_tag, qual_tag = extract_barcode(record, cb)
        try:
            corrected_cb = self._error_mapping.get_corrected_barcode(seq_tag[1])
        except KeyError:
            corrected_cb = seq_tag[1]  # couldn't find the right barcode!
        return seq_tag, qual_tag, ('CB', corrected_cb, 'Z')
```

This is bad, because it fails to separate categories of reads that should not be combined: 
Tags on read with correctible cell barcode: `CR`, `CB`, `CR` different from `CB`
Tags on read with perfect cell barcode: `CR`, `CB`, `CR` same as `CB`
Tags on read with nonsense cell barcode: `CR`, `CB`, `CR` same as `CB`

If instead we only store `CB` when we identify a perfect or correctible barcode, these three categories of reads would become separable. 

However, it would break some existing functionality. 

1. `SplitBamByCellBarcode`: currently we assign reads to file by the `CB` tag. This change would break this functionality. To fix this, we could default to `CR` in a `try-except` expression when `CB` is not found. 

2. `Metrics`: cell metrics aggregate information by cell barcode. reads without cell barcode would be excluded. This may be the optimal solution, since they definitively do not represent cells. Conversely, it would be important to know how many observations of genes occur in cells vs non-cells, so the number of reads aligning to genes but containing non-cell barcodes should be tracked, and should be comparable to the number of reads that do associate with known cell barcodes. 

3. Picard UmiAwareMarkDuplicates: I assume it will break because it's picard. Technically it doesn't use the CB tag, so we can probably remove it.

ambrosejcarr:
Looking for some feedback here on this problem + implementation proposal.

jsotobroad:
So I guess one worry I have for this is that this no one tag that someone can look at to know what is the "best" cell barcode for all reads in a bam which I thought was important for some reason (I could be totally off here.

My one question is what is the impact of having these barcodes that were not correctible with our current scheme still have a `CB` value.

2.  I think you could still go through the reads with "None" `CB` values and calculate some relevant metrics if you wanted to after you made this change
 
3. it actually currently wouldn't affect anything because it doesn't even use the tag like you pointed out.  Even when it is changed to use the CB tag to mark duplicates it should just ignore it if one is not present.

[...]

PR #18, opened 2018-03-19
Ajc fix cell barcode correction

ambrosejcarr:
* Fixes cell barcode correction as described in #17 
* Extracts a useful file type inference tool from `reader.py`
* defaults to gzip compressing metrics outputs


src/sctools/bam.py
```diff
@@ -187,20 +189,28 @@ def _cleanup(files_to_counts, files_to_names, rm_all=False):
...
+        # create an empty map for (tag, barcode) to files
         tags_to_files = {}
 
-        # loop over input, partitioning barcodes into files
+        # loop over input; check each tag in sequence and partition barcodes into files
         for alignment in input_alignments:
-            try:
-                tag_content = alignment.get_tag(tag)
-            except KeyError:
+            for tag in tags:
```
rexwangcc: This seems expensive to me, is this "check each tag in sequence and partition" avoidable? (Potentially we could write this into a `map()` function so in the future multiprocessing could be added here for efficiency-sensible tasks, ignore me if I'm wrong here)

src/sctools/reader.py
```diff
@@ -3,6 +3,35 @@
 import bz2
 from copy import copy
 from collections.abc import Iterable
+from functools import partial
+from typing import Callable
+
+
+def infer_open(file_: str, mode: str):
```
rexwangcc: How about `def infer_open(file_: str, mode: str) -> Callable:`?

src/sctools/bam.py
ambrosejcarr: You found a bug! It needed a `break` in the `try` block.

PR #19, opened 2018-03-22
fix writing of reads without gene or cell records

ambrosejcarr:
Can I get a rubber stamp here @jsotobroad ? 

The change that I made in the previous PR didn't account for how to write the record that has `None` instead of a `GE` tag. This PR fixes that by simply calling `repr()` on the index (gene or cell tag) prior to writing it. If the tag is a string, this operation does nothing.


ambrosejcarr:
Thanks @dshiga

PR #20, opened 2018-03-23
add CircleCI

ambrosejcarr:
I've written 142 tests. Add CI that runs these on commit.

### DIFF
--- a/src/sctools/test/characterize-gene-testing-data.ipynb
+++ b/src/sctools/test/characterize-gene-testing-data.ipynb
@@ -270,7 +270,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 32,
+   "execution_count": 9,
    "metadata": {
     "run_control": {
      "frozen": false,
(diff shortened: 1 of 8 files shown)


src/sctools/test/characterize-gene-testing-data.ipynb
```diff
@@ -1149,7 +1141,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.1"
+   "version": "3.6.4"
```
rexwangcc: Why is this inconsistent with the circle ci's python version(image: circleci/python:3.6.1)? (I don't know if these two versions are even related tho, just curious question)

ambrosejcarr: This is metadata from the ipython notebook. It has nothing to do with the CircleCI version, but rather reflects the fact that I most recently ran this on my desktop, which is using python 3.6.4, instead of my laptop, which is using python 3.6.1 👍

PR #27, opened 2018-03-30
Ajc docs

ambrosejcarr:
I did some off-board work update the docs. It needs doctest examples, but those will take time that I don't have right now, so they'll have to wait. 

@rexwangcc think you can take a look through this? I'd love to know what you think it's still missing. 

I think what will make sense is to identify points where improvement is possible and make issues, but merge this quickly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,7 +16,6 @@ Download and Installation
    pip3 install .
    pytest  # verify installation; run tests
 
-
 sctools Package
 ---------------
 
@@ -62,3 +61,17 @@ type the following two commands to generate the report and open it in your web b
 .. code bash
    pytest --cov-report html:cov_html --cov=sctools
    open cov_html/index.html
+
+Definitions
+-----------
+
+1. Fragment
+
+2. Molecule
+
+3. Cell Barcode
+
+4. Molecule Barcode
+
+5. Bam/Sam file: unless specified, refers to any aligned or unaligned bam/sam file
+# todo add more
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,7 @@ to calculate and view test coverage cd to the ``sctools`` directory and
 type the following two commands to generate the report and open it in your web browser:
 
 .. code:: bash
+
    pytest --cov-report html:cov_html --cov=sctools
    open cov_html/index.html
 
--- a/README.rst
+++ b/README.rst
@@ -77,32 +77,32 @@ Definitions
 
 Several definitions are helpful to understand how sequence data is analyzed.
 
-1. Cell: an individual cell, the target of single-cell RNA-seq experiments and the entity that we
+1. **Cell**: an individual cell, the target of single-cell RNA-seq experiments and the entity that we
 with to characterize
 
-2. Capture Primer: A DNA oligonucleotide containing amplification machinery, a fixed cell barcode,
+2. **Capture Primer**: A DNA oligonucleotide containing amplification machinery, a fixed cell barcode,
 a random molecule barcode, and an oligo-dT tail to capture poly-adenylated RNA
 
-3. Molecule: A molecule refers to a single mRNA molecule that is captured by an oligo-dT capture
+3. **Molecule**: A molecule refers to a single mRNA molecule that is captured by an oligo-dT capture
 primer in a single-cell sequencing experiment
 
-4. Molecule Barcode: A molecule barcode (alias: UMI, RMT) is a short, random DNA barcode attached
+4. **Molecule Barcode**: A molecule barcode (alias: UMI, RMT) is a short, random DNA barcode attached
 to the capture primer that has adequate length to be probabilistically unique across the experiment.
 Therefore, when multiple molecules of the same gene are captured in the same cell, they can be
 differentiated through having different molecule barcodes. The proposed GA4GH standard tag for a
 molecule barcode is UB and molecule barcode qualities is UY
 
-5. Cell Barcode: A short DNA barcode that is typically selected from a whitelist of barcodes that
+5. **Cell Barcode**: A short DNA barcode that is typically selected from a whitelist of barcodes that
 will be used in an experiment. All capture primers for a given cell will contain the same cell
 barcode. The proposed GA4GH standard tag for a cell barcode is CB and cell barcode qualities is CY
 
-6. Fragment: During library construction, mRNA molecules captured on capture primers are amplified,
+6. **Fragment**: During library construction, mRNA molecules captured on capture primers are amplified,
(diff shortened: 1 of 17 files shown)


rexwangcc:
These look really good to me, I will also look at them today later as off-board works. Will comment on them.

README.rst
```diff
@@ -53,12 +57,52 @@ Main Package Classes
...
 Viewing Test Results and Coverage
 ---------------------------------
+
 to calculate and view test coverage cd to the ``sctools`` directory and
 type the following two commands to generate the report and open it in your web browser:
 
 .. code bash
    pytest --cov-report html:cov_html --cov=sctools
    open cov_html/index.html
+
```
rexwangcc: It looks like the above block is not working in the Github rendered readme page?

README.rst
```diff
@@ -53,12 +57,52 @@ Main Package Classes
...
+6. **Fragment**: During library construction, mRNA molecules captured on capture primers are amplified,
+and the resulting amplified oligonucleotides are fragmented. In 3' experiments, only the fragment
+that contains the 3' end is retained, but the break point will be random, which means fragments
+often have different lengths. Once sequenced, different fragments can be identified as unique
+combinations of cell barcode, molecule barcode, the chromosome the sequence aligns to, and the
+position it aligns to on that chromosome, after correcting for clipping that the aligner may add
+
+7. **Bam/Sam file**: The GA4GH standard file type for the storage of aligned sequencing reads.
+Unless specified, our Single Cell Tools will operate over bam files containing either aligned or
+unaligned reads
```
rexwangcc: This readme is the best documentation among all of our repos!

src/sctools/bam.py
```diff
@@ -1,55 +1,107 @@
...
-          include_other is not 0.
+                raise ValueError(
+                    'could not autodetect file type for alignment_file %s (detectible suffixes: '
+                    '.sam, .bam)' % alignment_file
+                )
+        self._file: str = alignment_file
+        self._open_mode: str = open_mode
+
+    # todo figure out how to generate optional output type hints
+    def indices_by_chromosome(self, n_specific: int, chromosome: str, include_other: int=0):
```
rexwangcc: The inputs are using PEP484, I'd also make the return values comply with PEP484, i.e. `def indices_by_chromosome(self, n_specific: int, chromosome: str, include_other: int=0) -> ???` for this case. Although Numpy docstring force you to duplicate the return type in the docstring, I think adding `->` is better, from the "self-documented code" perspective. 

Besides, for the todo, I guess:
```python
from typing import Union

f(x) -> Union[int, None]:
```
could be a feasible solution here?

rexwangcc: Ah, after going through the implementation, I feel the annotations would be:

```python
from typing import Union, Tuple
def indices_by_chromosome(self, n_specific: int, chromosome: str, include_other: int=0) -> Union[Tuple[List[int], List[int]], List[int]]:
```
This seems to be more complicated now, a pain of Typing :(

src/sctools/bam.py
```diff
@@ -130,34 +193,56 @@ def tag(self, output_bam_name, tag_generators):
...
+        The bam tags to split on. The tags are checked in order, and sorting is done based on the
+        first identified tag. Further tags are only checked if the first tag is missing. This is
+        useful in cases where sorting is executed over a corrected barcode, but some records only
+        have a raw barcode.
+    approx_mb_per_split : float
+        The target file size for each chunk in mb
+    raise_missing : bool, optional
+        if True, raise a RuntimeError if a record is encountered without a tag. Else silently
+        discard the record (default = True)
+
```
rexwangcc: It would be better to have a consistent annotation among the repo, either PEP484 to all funcs, or not. But it can be gradually done, no need to be finished in this PR.

src/sctools/metrics/gatherer.py
```diff
@@ -1,41 +1,109 @@
...
 class GatherCellMetrics(MetricGatherer):
 
-    # assumes bam file is sorted by GE, UB, CB tags
+    extra_docs = """
+    Notes
+    -----
+    ``bam_file`` must be sorted by gene (``GE``), molecule (``UB``), and cell (``CB``), where gene 
+    varies fastest.
+    
+    Examples
```
rexwangcc: The examples here look really appealing, though they always cost a lot of time. Again, it can be done gradually in the future.

src/sctools/test/test_metrics.py
```diff
@@ -344,7 +344,7 @@ def test_higher_order_metrics_by_gene(metrics, key, expected_value):
     the test suite across all measured instances of the metric class. E.g. for cell metrics (class),
     each test will verify the value for each cell (instance).
 
-    :param pd.DataFrame metrics: Output from subclass of sctools.metrics.SequenceMetricAggregator
+    :param pd.DataFrame metrics: Output from subclass of sctools.metrics.MetricAggregator
```
rexwangcc: It seems this test file is still using reST style?

rexwangcc: I figure test cases should have minimal doc strings to save the time of developers. So I'm happy with it if you don't apply numpy style on these files.

rexwangcc:
Oops, it seems the `Codecov Report` agent is talking loudly here. It can be turned off through the codecov.io anytime.

src/sctools/metrics/gatherer.py
ambrosejcarr: Agree, I think the examples make the code base way more usable. They also let me avoid stupid things. For example, I sometimes feel compelled to write wrappers for existing functions to make usage obvious... examples can solve that problem.

src/sctools/bam.py
ambrosejcarr: Sorry, missed this one 👍

src/sctools/bam.py
ambrosejcarr: We agree on this implementation. Added it.

ambrosejcarr:
@rexwangcc Want a fun brain teaser? Figure out the return call of `sctools.reader.zip_readers`

ambrosejcarr:
>  Spinx + readthedocs hooks could be added to the docstrings during the development in the future.

Agree. I want to add this and a CONTRIBUTING.md file. I've already put in some of the sphinx hooks, but will need to figure out napolean to get the numpydoc to work with sphinx.

ambrosejcarr:
Thanks for all the comments. I think I fixed everything. Mind taking another look?

rexwangcc:
Sure, will do. For the brain teaser, what do think about `Generator[Tuple[Reader], None, None]`?

PR #34, opened 2018-04-11
ENH: add map-reduce count matrix construction with npy output

ambrosejcarr:
Adds two methods to construct and merge count matrices that currently output .npy, but generate a scipy.sparse.csr matrix internally. The intention is to switch this over to the format decided on by the Matrix Service prior to go-live. 

New Modules:
* CreateCountMatrix
* MergeCountMatrix


src/sctools/count.py
```diff
@@ -0,0 +1,163 @@
...
+    -------
+    count_matrix : sp.csr_matrix
+        cells x genes sparse count matrix in compressed sparse row format (cells are compressed)
+
+    Notes
+    -----
+    Any matrices produced by this function that share the same annotation file can be concatenated
+    using the scipy sparse vstack function, for example:
+
+    >>> import scipy.sparse as sp
```
rexwangcc: 👍

src/sctools/count.py
```diff
@@ -0,0 +1,163 @@
...
+    gene_indices: List[int] = []
+
+    gene_id_to_index: Dict[str, int] = {}
+    gtf_reader = gtf.Reader(annotation_file)
+
+    # map the gene from reach record to an index in the sparse matrix
+    for gene_index, record in enumerate(gtf_reader.filter(retain_types=['gene'])):
+        try:
+            gene_id = record.get_attribute('gene_id')
+        except KeyError:
```
rexwangcc: I have a feeling that you could only get a None instead of a KeyError from using this method: https://github.com/HumanCellAtlas/sctools/blob/53b22c935bd537ce901db541899ca7d5f249c01f/src/sctools/gtf.py#L146, since it is `dic.get('key')` instead of `dic['key']`. I could be wrong though.

src/sctools/platform.py
```diff
@@ -219,6 +224,105 @@ def merge_cell_metrics(cls, args: Iterable[str]=None) -> int:
...
+            value of None, when passed to `parser.parse_args` causes the parser to
+            read `sys.argv`
+
+        Returns
+        -------
+        return_call : 0
+            return call if the program completes successfully
+
+        """
+        parser = argparse.ArgumentParser()
```
rexwangcc: [Irrelevant] Recently I found a lib called [Click](http://click.pocoo.org/5/), which is developed by the author of flask and werkzeug. I found it really elegant and maybe in the future it can be used to replace the parsers here.

ambrosejcarr: Click is nice for simple parsers but it doesn't work for quite a few more complex things (including inheritance). I prefer `argparse`, but agree that I'm not yet using the `argparse` options that I'm alluding to because we haven't expanded the library to have multiple different 3' pipelines. 

When we have nuclei, in-drop, and drop-seq, I suspect we'll want `argparse`. If that's wrong I will revisit. :)

src/sctools/count.py
ambrosejcarr: Nice 🐛

rexwangcc:
Oops, I meant comment not explicit 👍

ambrosejcarr:
Hey Jishu, 

Thanks for the comments. Addressing your questions: 

1. The outputs are a compressed sparse row matrix, currently saved in numpy binary for easy of development. This will be changed to whatever the matrix service outputs. 

For the test data, it's an empty matrix. For larger datasets, it looks better, although I haven't investigated whether it looks like "valid biology" -- I'm waiting for the test data for that. 

2. Multi-aligned reads should output single gene ids when they get resolved or concatenated gene ids when they don't. This matrix includes both single and concatenated gene ids (whatever is in the GE tag) without any loss of efficiency, since the output is sparse. That said, I have **not** tested this because the multi-alignment tool isn't integrated into skylab yet. 

I don't see any reason why it shouldn't work, but if this breaks anything it shouldn't be a hard fix. 

Could you please take a look at the skylab PR as well? that one needs a CB review to merge. Thanks!

PR #37, opened 2018-04-21
[Experimental] Attempt to add automated versioning.

rexwangcc:
This PR is intending to add automated versioning provided by [`setuptools_scm`](https://github.com/pypa/setuptools_scm), so that we don't need to manually maintain `setup.py` and `conf.py`, which is painful and trivial.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,6 +14,7 @@
 #
 import os
 import sys
+from pkg_resources import get_distribution
 sys.path.insert(0, os.path.abspath('..'))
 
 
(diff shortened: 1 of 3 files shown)


ambrosejcarr:
Seems like magic. How does it work?

rexwangcc:
@ambrosejcarr I haven't looked into how does it work internally, but it will try to get the version in the order described here: https://github.com/pypa/setuptools_scm#default-versioning-scheme

rexwangcc:
@ambrosejcarr Are you up for this? If not, I'm going to close this PR

rexwangcc:
@ambrosejcarr This package works for installing from both PyPi and Github, but the prerequisite here is that you need to host your source code on Git-based platforms(Github, GitLab), and since at this moment, sctools is already on Github, I think it is fine.

I must say I haven't understood the full logic of the setuptools_scm, but I think it is worth trying it. And we can just remove it if there is anything bad in the future during the development process. What do you think?

ambrosejcarr:
👍

PR #43, opened 2018-06-27
CountMatrix bugfix and new tests

### LINKED ISSUE #42: Fix and improve the test case for CountMatrix
As it stands, the test is wrong and the design needs to be improved.

A good starting point is to generate a synthetic test BAM file on the fly. Note that only the TAGs (and potentially FLAGs) matter, so the template sequences and qualities can all be set to `N`.

- For a subset of genes, we generate only unique alignments.
- For another subset, we generate multiple alignments.
- We also through in a number of records with missing `CB` or `UB`.
- We generate the BAM file in the expected sort order, such that dependence on `bamtools` would not be required for testing.

By construction, we _know_ what count matrix we expect to get. Hence, this would be a good test for `CountMatrix.from_bam`.

mbabadi:
This PR fixes some algorithmic issues in the `CountMatrix` class. It also improves related unit tests. The current counting algorithm implemented in `CountMatrix.from_sorted_tagged_bam` is expected to be functionally similar to CellRanger methods.

Closes #42.


src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+
+        # generate gene names
+        self.all_gene_names = self._load_gene_names(annotation_file)
+        self.max_genes = max_genes
+        self.num_genes = len(self.all_gene_names)
+        self.to_be_used_gene_indices: List[int] = self.rng.choice(
+            np.arange(0, self.num_genes, dtype=np.int), size=self.max_genes, replace=False).tolist()
+        self.to_be_used_gene_names = [self.all_gene_names[j] for j in self.to_be_used_gene_indices]
+
+    def generate_synthetic_counts_and_alignment_tags(self,
```
ambrosejcarr: I like parameter lists to go on the next line to save on leading white space. Also because it always works--in cases where you have really long function + parameter names the style used here can overrun your line limit.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+                                                     num_missing_some_tags: int,
+                                                     num_multiple_gene_alignments: int,
+                                                     max_gene_hits_per_multiple_gene_alignments: int)\
+            -> SyntheticDataBundle:
+
+        # generate count matrix
+        count_matrix: np.ndarray = self._generate_random_count_matrix()
+
+        # generate primary alignment tags that produce count_matrix
+        primary_alignment_record_tags_set, row_index, col_index = \
```
ambrosejcarr: instead of a forward slash, I believe you can use an open parenthesis. [PEP8/maximum-line-length](https://www.python.org/dev/peps/pep-0008/?#maximum-line-length) (see bottom of section).

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+            self._generate_primary_alignment_record_bundle(count_matrix)
+        primary_alignment_record_tags_list = list(primary_alignment_record_tags_set)
+
+        # sanity check -- we require as many primary alignment records as the total counts
+        assert len(primary_alignment_record_tags_set) == np.sum(count_matrix)
```
ambrosejcarr: I would prefer the comment to be expressed as the failure string, so users breaking this know exactly what's gone wrong: 
```python
        assert len(primary_alignment_record_tags_set) == np.sum(count_matrix), \
            'We require as many primary alignment records as the total counts'
```

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        incomplete_alignment_tags_list: List[AlignmentRecordTags] = self._generate_incomplete_alignment_tags(
+            num_missing_some_tags)
+
+        # add records with multiple gene alignments
+        multiple_alignment_tags_list: List[List[AlignmentRecordTags]] = self._generate_multiple_gene_alignment_tags(
+            num_multiple_gene_alignments,
+            max_gene_hits_per_multiple_gene_alignments,
+            primary_alignment_record_tags_set)
+
+        return SyntheticDataBundle(count_matrix,
```
ambrosejcarr: Similar to the function definitions, I'd put the first argument on the newline, indented only once from the enclosing scope.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+                                   col_index,
+                                   primary_alignment_record_tags_list,
+                                   duplicate_alignment_tags_list,
+                                   incomplete_alignment_tags_list,
+                                   multiple_alignment_tags_list)
+
+    def generate_synthetic_bam_and_counts_matrix(self,
+                                                 output_path: str,
+                                                 num_duplicates: int,
+                                                 num_missing_some_tags: int,
```
ambrosejcarr: Can we add a guarantee that each tag is missed at least once for testing purposes, or split this out into three parameters, one per missing tag?

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        query_sort_order : str
+            'unsorted' : shuffle the records before saving to BAM
+            'CB_UB_GE_n' : sort hierarchically by CB, UB, GE, and query_name
+            'n' : only sort by query_name
+
+        Returns
+        -------
+        None
+        """
+        assert 2 <= max_gene_hits_per_multiple_gene_alignments < self.max_genes
```
ambrosejcarr: What do you think about adding failure strings that describe why each assertion is important? I haven't done this in other functions but I think it's good practice that I'd like to add here where possible.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+                    alignments.add(unique_alignment_tag)
+
+        return alignments, row_index, col_index
+
+    def _generate_unique_random_alignment_tag(self,
+                                              existing_alignment_tags: Set[AlignmentRecordTags],
+                                              gene_name: str,
+                                              cell_barcode: Optional[str] = None,
+                                              molecule_barcode: Optional[str] = None) -> AlignmentRecordTags:
+        assert gene_name in self.to_be_used_gene_names, f"{gene_name} is not an allowed gene for generating " \
```
ambrosejcarr: I think I'd put the slash here before the string, to give more room to expand it in the future.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+            number of duplicate records
+        num_missing_some_tags : int
+            number of records that miss at least one crucial tag
+        num_multiple_gene_alignments : int
+            number of records that have at least two different gene tags
+        max_gene_hits_per_multiple_gene_alignments : int
+            maximum number of unique gene names to use for multiple-gene records
+        query_sort_order : str
+            'unsorted' : shuffle the records before saving to BAM
+            'CB_UB_GE_n' : sort hierarchically by CB, UB, GE, and query_name
```
ambrosejcarr: What do you think about making this a `tuple`, and adding an `Enum` class somewhere that lists the valid sort fields in case we want the bam sorted in a different order? The gene metrics, for one, request a different sort order.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+                                 reference_lengths=[SyntheticDataBundle.synthetic_sequence_length]) as bo:
+            for record in records:
+                bo.write(record)
+
+        # write count matrix, row index, and col index
+        np.save(os.path.join(output_path, self.count_matrix_output_filename), synthetic_data_bundle.count_matrix)
+        np.save(os.path.join(output_path, self.row_index_output_filename), synthetic_data_bundle.row_index)
+        np.save(os.path.join(output_path, self.col_index_output_filename), synthetic_data_bundle.col_index)
+
+    def _generate_random_count_matrix(self) -> np.ndarray:
```
ambrosejcarr: The code here is clear but I think a doc block here and for the below functions (for developers) would be nice. What do you think?

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+
+        return SyntheticDataBundle(count_matrix,
+                                   row_index,
+                                   col_index,
+                                   primary_alignment_record_tags_list,
+                                   duplicate_alignment_tags_list,
+                                   incomplete_alignment_tags_list,
+                                   multiple_alignment_tags_list)
+
+    def generate_synthetic_bam_and_counts_matrix(self,
```
ambrosejcarr: Question: for python code, I like to read top to bottom. This function calls a bunch of the private methods below. My tendency would be to move this below those. Do you have any strong opinions on this?

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+    def _generate_duplicate_alignment_tags(self,
+                                           num_duplicates: int,
+                                           primary_alignments_list: List[AlignmentRecordTags])\
+            -> List[AlignmentRecordTags]:
+        return self.rng.choice(primary_alignments_list, size=num_duplicates).tolist()
+
+    def _generate_incomplete_alignment_tags(self, num_missing_some_tags: int) -> List[AlignmentRecordTags]:
+        incomplete_alignment_tags_list: List[AlignmentRecordTags] = list()
+        for _ in range(num_missing_some_tags):
+            tag_mask = self.rng.randint(low=0, high=7)
```
ambrosejcarr: Nice, I like this way of masking. Very efficient!

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+@pytest.mark.parametrize('query_sort_order', ['CB_UB_GE_n', 'n'])
+def test_count_matrix_from_bam(query_sort_order):
+    # instantiate a test data generator
+    synthetic_data_generator = SyntheticBarcodedBAMGenerator(
+        _test_num_cells,
+        _test_max_genes,
+        _test_annotation_file,
+        _test_gene_expression_rate)
+
+    with tempfile.TemporaryDirectory() as _test_temp_dir:
```
ambrosejcarr: I'd add a `finally` clause to explicitly `shutil.rmtree` the temp dir -- on some operating systems, these files get left around until restart, so it's nice to clean up after ourselves. :)

src/sctools/test/test_metrics.py
```diff
@@ -1,13 +1,12 @@
+import fileinput
```
ambrosejcarr: Thanks for sorting 👍

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        count_matrix_data_expected, row_index_expected, col_index_expected)
 
-    # check matrices are the same
-    csr1: sp.csr_matrix = count_matrix._matrix
-    csr2: sp.csr_matrix = count_matrix_2._matrix
-    assert np.allclose(csr1.indices, csr2.indices)
-    assert np.allclose(csr1.indptr, csr2.indptr)
-    assert np.allclose(csr1.data, csr2.data)
+    # assert equality
+    assert np.allclose(sorted_count_matrix_data_from_bam, sorted_count_matrix_data_expected)
```
ambrosejcarr: Can we add a docstring to this test that's super clear on what we're testing here? My initial concern was that we're generating synthetic bams and count matrices, and using similar code to transition between them, instead of using some kind of external gold standard that we're aiming to hit. 

Done well, these tests are probably way better than some small matrix or bam used to smoke test the functionality, but there's also an opportunity to code `2 + 2 = 5` somewhere into both the count matrix generation and the bam_to_count function. 

[moonshot] Longer term, when we've figured out all the discrepancies, what do you think about trying to run the cell ranger code in our tests to verify we're equivalent?

src/sctools/count.py
```diff
@@ -40,20 +40,101 @@ def __init__(self, matrix: sp.csr_matrix, row_index: np.ndarray, col_index: np.n
...
+    def from_sorted_tagged_bam(
             cls,
             bam_file: str,
             annotation_file: str,
             cell_barcode_tag: str='CB',
             molecule_barcode_tag: str='UB',
-            gene_id_tag: str='GE',
-            open_mode: str='rb',
-            ):
+            gene_name_tag: str='GE',
```
ambrosejcarr: One question I forgot to ask: are these HGNC gene names that are going in the tags, and not the numerical IDs? If so then this is a good change. 

If they are IDs, I'd prefer to keep the name as ID.

src/sctools/count.py
```diff
@@ -40,20 +40,101 @@ def __init__(self, matrix: sp.csr_matrix, row_index: np.ndarray, col_index: np.n
...
+    @property
+    def row_index(self):
+        return self._row_index
+
+    @property
+    def col_index(self):
+        return self._col_index
+
+    @staticmethod
+    def _get_alignments_grouped_by_query_name_generator(bam_file: str,
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/count.py
```diff
@@ -40,20 +40,101 @@ def __init__(self, matrix: sp.csr_matrix, row_index: np.ndarray, col_index: np.n
...
+        molecule_barcode_tag : str
+            Tag that specifies the molecule barcode for each read.
+
+        Returns
+        -------
+            a generator for tuples (query_name, cell_barcode, molecule_barcode, alignments)
+        """
+        with pysam.AlignmentFile(bam_file, mode=open_mode) as bam_records:
+            for alignment in itertools.groupby(bam_records, key=lambda record: record.query_name):
+                query_name: str = alignment[0]
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/gtf.py
```diff
@@ -244,3 +247,44 @@ def filter(self, retain_types: Iterable[str]) -> Generator:
...
+    Dict[str, int]
+        A map from gene names to their linear index
+    """
+    gene_name_to_index: Dict[str, int] = dict()
+    gene_index = 0
+    for record in Reader(files, mode, header_comment_char).filter(retain_types=['gene']):
+        gene_name = record.get_attribute('gene_name')
+        if gene_name is None:
+            raise ValueError(
+                'Malformed GTF file detected. Record is of type gene but does not have a '
```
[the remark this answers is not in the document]
mbabadi: Agreed, and done.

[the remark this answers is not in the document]
mbabadi: While at it, I converted all exception messages to f-stye in the repo.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+    sorted_row_indices = [idx for idx, _ in sorted(enumerate(row_index), key=operator.itemgetter(1))]
+    sorted_col_indices = [idx for idx, _ in sorted(enumerate(col_index), key=operator.itemgetter(1))]
+    return (count_matrix[sorted_row_indices, :][:, sorted_col_indices],
+            row_index[sorted_row_indices],
+            col_index[sorted_col_indices])
+
+
+@pytest.mark.parametrize('query_sort_order', ['CB_UB_GE_n', 'n'])
+def test_count_matrix_from_bam(query_sort_order):
```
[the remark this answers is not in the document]
mbabadi: Nice suggestion, but let me pass on this one -- it's too much work for little gain. The main speed bottleneck in this test, by and large, is loading chr1 annotations. By the way, it is not uncommon for test suites to take hours to complete! (GATK takes about 1 hour).

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+_test_num_cells = 100
+_test_max_genes = 50
+_test_gene_expression_rate = 5.0
+_test_num_duplicates = 100
+_test_num_missing_some_tags = 100
+_test_num_multiple_gene_alignments = 100
+_test_max_gene_hits_per_multiple_gene_alignments = 10
+
+
+class AlignmentRecordTags:
```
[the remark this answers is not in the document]
mbabadi: Nice suggestion, however, there's a bizarre bug in `numpy.random.choice` that ruins the game. We can not choose randomly from lists of tuples! This is due to an unexpected casting of the list to `ndarray` which results in casting a `List[NamedTuple]` of size `N` to a `N x k` matrix where `k` is the number of elements in the tuple.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+
+    See Also
+    --------
+    SyntheticBarcodedBAMGenerator
+
+    """
+
+    synthetic_sequence_name = "SYNTHETIC_SEQUENCE"
+    synthetic_sequence_length = 100
+    _primary_query_name_prefix = "PRIMARY_QUERY_"
```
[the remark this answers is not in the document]
mbabadi: To denote internal use (PEP8). Python has no explicit public/private/protected modifiers. It is good practice to use `_var_name` for intended private (and protected) variables.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+    _duplicate_query_name_prefix = "DUPLICATE_QUERY_"
+    _incomplete_query_name_prefix = "INCOMPLETE_QUERY_"
+    _multi_gene_query_name_prefix = "MULTI_GENE_QUERY_"
+
+    def __init__(self,
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+_test_gene_expression_rate = 5.0
+_test_num_duplicates = 100
+_test_num_missing_some_tags = 100
+_test_num_multiple_gene_alignments = 100
+_test_max_gene_hits_per_multiple_gene_alignments = 10
+
+
+class AlignmentRecordTags:
+    """Stores cell barcode, molecule barcode, and gene name."""
+    def __init__(self,
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+    _multi_gene_query_name_prefix = "MULTI_GENE_QUERY_"
+
+    def __init__(self,
+                 count_matrix: np.ndarray,
+                 row_index: List[str],
+                 col_index: List[str],
+                 primary_alignment_record_tags_list: List[AlignmentRecordTags],
+                 duplicate_alignment_tags_list: List[AlignmentRecordTags],
+                 incomplete_alignment_tags_list: List[AlignmentRecordTags],
+                 multiple_alignment_tags_list: List[List[AlignmentRecordTags]]):
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        self.duplicate_alignment_tags_list = duplicate_alignment_tags_list
+        self.incomplete_alignment_tags_list = incomplete_alignment_tags_list
+        self.multiple_alignment_tags_list = multiple_alignment_tags_list
+
+        self.num_queries = (len(primary_alignment_record_tags_list)
+                            + len(duplicate_alignment_tags_list)
+                            + len(incomplete_alignment_tags_list)
+                            + len(multiple_alignment_tags_list))
+
+    def _get_query_name(self, query_prefix: str, i_query: int) -> str:
```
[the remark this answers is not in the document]
mbabadi: Added docstring: `Returns query name string from query index. We zero-pad the string representation of query indices merely for pretty-printing, e.g. 0000, 0001, ..., 9999.`

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        num_digits = len(str(self.num_queries - 1))
+        return query_prefix + str(i_query).zfill(num_digits)
+
+    def _generate_aligned_segment_from_tags(self,
+                                            alignment_tags: AlignmentRecordTags,
+                                            reference_start: int,
+                                            query_prefix: str,
+                                            i_query: int) -> pysam.AlignedSegment:
+        """Generates pysam.AlignedSegment instances from the contained tags.
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+
+    def get_bam_records_generator(self, rng_seed: int = 777) -> Generator[pysam.AlignedSegment, None, None]:
+        """Returns a generator of pysam.AlignedSegment instances created from the alignment tags
+        provided to the initializer."""
+        rng = np.random.RandomState(rng_seed)
+        i_query = 0
+
+        # primary alignments
+        for alignment_tags in self.primary_alignment_record_tags_list:
+            reference_start = rng.randint(low=0, high=self.synthetic_sequence_length)
```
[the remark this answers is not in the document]
mbabadi: At the moment, `sequence_name` and `sequence_start` are completely random and arbitrary (since we are not using these alignment features for counting -- only tags). Perhaps it makes sense if we go ahead and generate appropriate BAM header from hg38 reference and `sequence_name` and `sequence_start` and CIGAR strings that actually to an alignment to the gene shown in the tag. I would, however, willing to do this if we plan to feed the BAM file to cellranger (or use these features in counting). Let's discuss.

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        generates synthetic test data as described above and returns an instance of SyntheticDataBundle
+    generate_synthetic_bam_and_counts_matrix
+        generates synthetic test data and writes them to disk
+
+    See Also
+    --------
+    count.from_sorted_tagged_bam
+    """
+
+    cell_barcode_tag = 'CB'
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
mbabadi: I will add additional docblocks regarding the nature of the test data.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
mbabadi: Also, had we used a truncated (or synthetic) set of annotations, we would not have discovered the indexing bug caused by multiple entries per gene!

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+        i_query = 0
+
+        # primary alignments
+        for alignment_tags in self.primary_alignment_record_tags_list:
+            reference_start = rng.randint(low=0, high=self.synthetic_sequence_length)
+            yield self._generate_aligned_segment_from_tags(
+                alignment_tags, reference_start, self._primary_query_name_prefix, i_query)
+            i_query += 1
+
+        # duplicate alignments
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
mbabadi: I also made an issue (#44) to further refactor out all constants in sctools to a single 'consts.py' file and took the first step in doing so.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
mbabadi: (Also, refactored out record generator methods from `SyntheticDataBundle` to `SyntheticBarcodedBAMGenerator` where they belonged)

src/sctools/test/test_count.py
```diff
@@ -1,44 +1,578 @@
...
+    cell_barcode_tag = 'CB'
+    molecule_barcode_tag = 'UB'
+    gene_name_tag = 'GE'
+    output_prefix = 'synthetic_'
+    bam_output_filename = output_prefix + 'records.bam'
+    count_matrix_output_filename = output_prefix + 'count_matrix.npy'
+    row_index_output_filename = output_prefix + '_row_index.npy'
+    col_index_output_filename = output_prefix + '_col_index.npy'
+
+    def __init__(self,
```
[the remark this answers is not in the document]
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: I added `AlignmentSortOrder` base class and two instances to `bam.py`.

src/sctools/test/test_count.py
mbabadi: Done.

src/sctools/test/test_count.py
mbabadi: I definitely agree -- I moved things around for better code readability.

src/sctools/test/test_count.py
mbabadi: :)

src/sctools/test/test_count.py
mbabadi: I did it just to be safe, but I have a feeling that this is redundant. There are several instances of improper usage of `tempfile.TemporaryDirectory()` in `sctools` and that is likely to be the cause of improper cleanup.

src/sctools/count.py
mbabadi: Unfortunately, they actually gene names and not IDs (we reply on `dropseq-tools` for tagging genes; it uses gene names by default, there is no way to force it to use gene IDs).

src/sctools/test/test_count.py
mbabadi: Added an extensive preamble documentation block.

mbabadi:
@ambrosejcarr I'm done -- thanks for the thorough review! let me rebase and regroups the commits and it will be good to go for your next review.

mbabadi:
@ambrosejcarr back to you.

mbabadi:
@ambrosejcarr please feel free to correct my grammar in the docs (I am very sloppy with plain language, specially while coding!) + stylistic inconsistencies -- I could notice a few typos/grammatical issues while browsing the diff! :)

src/sctools/test/test_count.py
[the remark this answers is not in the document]
ambrosejcarr: Sigh. Ducktyping. :-(

Thanks for testing it out.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
ambrosejcarr: I believe pretty strongly in responsive unit testing -- it aids in a fast development cycle. Mocks and fixtures can help with this, and they are not difficult to work with once you get a hang of it. I would bet that @rexwangcc could definitely help with this. 

Given what you said about loading the annotations, I'd suggest cutting down on the chr1 annotation sizes, and loading the _annotations_ as a fixture so they only get loaded once for further tests.

Could you use a smaller subset of the annotations but still maintain the same testing quality? This repo is not a great example, as we obviously have testing holes, but the existing tests complete in < 5s, so I think there's a burden to prove that the tests you're adding need to be so much long er.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
ambrosejcarr: Got it. Thanks.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
ambrosejcarr: Ah got it, you're query name sorting, so you don't need the alignment information? 

One part of me wishes we'd just fill with 0/nan/none/NNNN or something very obviously unused, but I can see the advantage of putting the right machinery in place to generate real sequences for more robust testing later. 

More docblocks seems like the right approach here. 👍

src/sctools/test/test_count.py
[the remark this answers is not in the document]
ambrosejcarr: Awesome. Definitely a big improvement. Thanks for the issue.

src/sctools/test/test_count.py
ambrosejcarr: Cool, made #45.

src/sctools/bam.py
```diff
@@ -231,27 +236,27 @@ def split(in_bam, out_prefix, *tags, approx_mb_per_split=1000, raise_missing=Tru
     if len(tags) == 0:
         raise ValueError('At least one tag must be passed')
 
-    def _cleanup(files_to_counts, files_to_names, rm_all=False) -> None:
+    def _cleanup(_files_to_counts, _files_to_names, rm_all=False) -> None:
```
ambrosejcarr: These parameters should have types.

ambrosejcarr: I am unclear on why you'd have a leading underscore here. It seems obvious that these methods are private since it's a function with no internal closures.

src/sctools/bam.py
```diff
@@ -260,8 +265,8 @@ def _cleanup(files_to_counts, files_to_names, rm_all=False) -> None:
         warnings.warn('Number of requested subfiles (%d) exceeds 500; this may cause OS errors by '
                       'exceeding fid limits' % n_subfiles)
     if n_subfiles > 1000:
-        raise ValueError('Number of requested subfiles (%d) exceeds 1000; this will usually cause '
-                         'OS errors, think about increasing max_mb_per_split.' % n_subfiles)
+        raise ValueError(f'Number of requested subfiles ({ n_subfiles}) exceeds 1000; this will usually cause '
```
ambrosejcarr: Why the leading space before `n_subfiles`?

src/sctools/bam.py
```diff
@@ -379,12 +383,12 @@ def iter_molecule_barcodes(bam_iterator: Iterator[pysam.AlignedSegment]) -> Gene
...
     ------
     grouped_by_tag : Iterator[pysam.AlignedSegment]
-        reads sharing a unique molecule barcode (``UB`` tag)
+        reads sharing a unique molecule barcode tag
     current_tag : str
         the molecule barcode that records in the group all share
 
     """
-    return iter_tag_groups(tag='UB', bam_iterator=bam_iterator)
+    return iter_tag_groups(tag=consts.MOLECULE_BARCODE_TAG_KEY, bam_iterator=bam_iterator)
```
ambrosejcarr: Nice 👍

src/sctools/bam.py
```diff
@@ -417,9 +421,69 @@ def iter_genes(bam_iterator: Iterator[pysam.AlignedSegment]) -> Generator:
     Yields
     ------
     grouped_by_tag : Iterator[pysam.AlignedSegment]
-        reads sharing a unique gene id (``GE`` tag)
+        reads sharing a unique gene name tag
```
ambrosejcarr: Yikes I thought these were gene ids, not gene names. Good catch.

src/sctools/gtf.py
```diff
@@ -244,3 +244,44 @@ def filter(self, retain_types: Iterable[str]) -> Generator:
...
+
+
+# todo this lenient behavior is deemed to change in the future (warning -> exception)
+def _resolve_multiple_gene_names(gene_name: str):
+    _logger.warning(f'Multiple entries encountered for "{gene_name}". Please validate the input GTF file(s). '
+                    f'Skipping the record for now; in the future, this will be considered as a '
+                    f'malformed GTF file.')
+
+
+def extract_gene_names(files='-', mode='r', header_comment_char='#') -> Dict[str, int]:
```
ambrosejcarr: What are your feeling on types for parameters with default arguments? Mypy suggests it because it enforces static typing on them (in addition to annotating them as strings, which is obvious given the default). I think I like this, but I'm open to opinions.

src/sctools/bam.py
mbabadi: Added types.

mbabadi: `files_to_counts` and `files_to_names` shadow from the outer scope. Even though this specific is OK, it is generally not a good practice to let inner classes use the outer scope var names (it is a recipe for hard-to-catch bugs!).

src/sctools/bam.py
mbabadi: Fixed.

mbabadi: Also, extracted constants.

src/sctools/bam.py
mbabadi: ;-)

src/sctools/gtf.py
mbabadi: Added types.

src/sctools/test/test_count.py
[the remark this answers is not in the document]
mbabadi: Alright, truncated chr1 annotations, changed some signatures, added gene names as a fixture. Test is fast now.

mbabadi:
@ambrosejcarr back to you.

src/sctools/bam.py
ambrosejcarr: Yep, agree. For some reason I'd extrapolated in my mind such that trailing underscores were used, where as weakly private methods have leading underscores. 

```
single_trailing_underscore_: used by convention to avoid conflicts with Python keyword, e.g.

Tkinter.Toplevel(master, class_='ClassName')
```

However, I couldn't find it in pep8.

PR #46, opened 2018-08-14
Add ability to sort bams by list of tags plus queryname

dshiga:
This will make it easier to sort BAMs using multiple tags, which in turn will facilitate calculating metrics.

See https://app.zenhub.com/workspace/o/humancellatlas/secondary-analysis/issues/170

Changes:
- Adds TagSortBam command and corresponding tag_sort_bam function in `platform.py`.
- Creates ComparableRecord class that wraps pysam.AlignedSegment and makes it easy to sort by tags and query name.
- Moves CellGeneMoleculeQueryNameSortOrder to test_count.py, since that's the only place it's used.

See related: https://github.com/HumanCellAtlas/skylab/pull/125


src/sctools/bam.py
```diff
@@ -464,7 +493,26 @@ def __repr__(self) -> str:
...
+
+    @property
+    def key_generator(self):
+        return self._get_sort_key
+
+    def __repr__(self) -> str:
+        return 'hierarchical__tags_query_name'
+
+
+class CellMoleculeGeneQueryNameSortOrder(TagsQueryNameSortOrder):
```
samanehsan: Is this class needed anymore now that there is a more generalized `TagsQueryNameSortOrder` class? It looks like it's only being used in some tests

src/sctools/bam.py
```diff
@@ -464,7 +493,26 @@ def __repr__(self) -> str:
         return 'query_name'
 
 
-class CellMoleculeGeneQueryNameSortOrder(AlignmentSortOrder):
+class TagsQueryNameSortOrder(AlignmentSortOrder):
```
samanehsan: Please update the docstring at the top of the file: https://github.com/HumanCellAtlas/sctools/blob/c2c2540801cf12d2f2fba72b2dabc6e82552a3d9/src/sctools/bam.py#L27

src/sctools/platform.py
```diff
@@ -44,6 +47,44 @@ class GenericPlatform:
...
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            '-i', '--input_bam', required=True,
+            help='input bamfile')
+        parser.add_argument(
+            '-o', '--output_bam', required=True,
+            help='output bamfile')
+        parser.add_argument('-t', '--tags', nargs='+',
+                            help='tag(s) to sort by')
+        parser.set_defaults(raise_missing=True)
```
samanehsan: Can you clarify what the `raise_missing` flag does here?

dshiga: Copy paste error - will remove.

src/sctools/test/test_bam.py
```diff
@@ -157,3 +157,41 @@ def test_split_succeeds_with_raise_missing_false_and_no_cr_barcode_passed(tagged
...
+    old_qn = ''
+    for record in sorted_records:
+        ub = bam.get_tag_or_default(record, 'UB')
+        cb = bam.get_tag_or_default(record, 'CB')
+        ge = bam.get_tag_or_default(record, 'GE')
+        qn = record.query_name
+
+        assert ub >= old_ub
+        if ub > old_ub:
+            cb = ''
```
samanehsan: Can you clarify why these tags get set to `''`? If the next record has the same `UB` but the `CB` is not in the right order would this throw off the comparison?

src/sctools/bam.py
dshiga: @ambrosejcarr @mbabadi if you want to keep the `CellMoleculeGeneQueryNameSortOrder` class for any reason let me know. Otherwise I will take it out, now that the more generic `TagQueryNameSortOrder` exists.

samanehsan:
Do the codecov checks fail when there's a decrease in test coverage? I've never used it before!

src/sctools/test/test_bam.py
dshiga: If ub has changed, it means cb can be anything and still be in the correct order. We set cb/old_cb to `''` so that for the next line any value of cb will be accepted. Will add some commenting to the code.

```
ub: 1 cb: 1
ub: 1 cb: 2
ub: 2 cb: 1
ub: 2 cb: 2
ub: 3 cb: 1
```

samanehsan: Ok, that part makes sense! I was thinking of how to check for the situation where the ub doesn't change and the cb is not in the right order:
```
ub: 1 cb: 1
ub: 1 cb: 2
ub: 2 cb: 2  
ub: 2 cb: 1 <- cb sort order is incorrect
ub: 3 cb: 1
```

dshiga: Yeah, so that's why we check that ub has changed in that first conditional on 178 and then only set cb to `''` if ub has changed.

dshiga: Thinking about this some more, you are totally right! There is a bug here. Going to revamp this.

src/sctools/bam.py
ambrosejcarr: I'm fine with taking it out.

src/sctools/bam.py
```diff
@@ -464,29 +500,83 @@ def __repr__(self) -> str:
...
+            tag_value = get_tag_or_default(record, key, '')
+            tag_values.append(tag_value)
+        query_name = record.query_name
+        return cls(tag_keys, tag_values, query_name, record)
+
+    def __lt__(self, other: 'TagSortableRecord'):
+        self.__verify_tag_keys_match(other)
+        for i in range(len(self.tag_keys)):
+            if self.tag_values[i] < other.tag_values[i]:
+                return True
```
samanehsan: Are both of these conditions necessary? This could be condensed into `return self.tag_values[i] < other.tag_values[i]`

src/sctools/bam.py
```diff
@@ -464,29 +500,83 @@ def __repr__(self) -> str:
...
+                raise ValueError('AlignedSegment is missing')
+            aligned_segments.append(cr.record)
+        return aligned_segments
 
-    def __repr__(self) -> str:
-        return 'hierarchical__cell_molecule_gene_query_name'
+    @staticmethod
+    def verify_sort(records: Iterable[pysam.AlignedSegment], tag_keys: List[str]) -> None:
+        """Raise AssertionError if the given records are not correctly sorted by the given tags and query name"""
+        old_record = SortFirstTagSortableRecord(tag_keys=tag_keys)
```
samanehsan: I think it would be clearer to instantiate a `TagSortableRecord` here directly: `old_record = TagSortableRecord(tag_keys=tag_keys, tag_values=[''] * len(tag_keys), query_name='', record=None)` instead of abstracting it into another class.

src/sctools/test/test_count.py
```diff
@@ -100,6 +100,34 @@ def __repr__(self):
                f'{consts.GENE_NAME_TAG_KEY}: {self.gene_name}'
 
 
+class CellMoleculeGeneQueryNameSortOrder(bam.AlignmentSortOrder):
```
samanehsan: I think this class should be in the same file as `AlignmentSortOrder` and `QueryNameSortOrder()`, whether that's here or in `bam.py`

src/sctools/test/test_bam.py
```diff
@@ -157,3 +194,71 @@ def test_split_succeeds_with_raise_missing_false_and_no_cr_barcode_passed(tagged
...
+    for i in range(len(SORTED_VALUES)):
+        r = bam.TagSortableRecord(TAGS, SORTED_VALUES[i][0], SORTED_VALUES[i][1])
+        records.append(r)
+    bam.TagSortableRecord.verify_sort(records, TAGS)
+
+
+def test_tags_queryname_sort_order_gives_correct_order():
+    tags = ['UB', 'CB', 'GE']
+    _, records = bam.read_bam(data_dir + 'queryname-sorted.bam')
+    sorted_records = bam.sort_by_tags_and_queryname(records, tags)
```
samanehsan: Can you also add a test for `sort_by_tags_and_queryname` where `tags` is None?

src/sctools/test/test_bam.py
```diff
@@ -157,3 +194,71 @@ def test_split_succeeds_with_raise_missing_false_and_no_cr_barcode_passed(tagged
     os.remove(tagged_bam)  # clean up
     for f in glob.glob('test_output_*'):
         os.remove(f)
+
+
+def test_tag_sortable_records_sort_correctly():
```
samanehsan: I find this test name slightly confusing since it's not sorting the records. Maybe rename this to `test_tag_sortable_record_comparison`?

samanehsan: I think having more specific unit tests can help with debugging if there are breaking changes, e.g. a test that just checks that two records with different query names are sorted correctly.

src/sctools/bam.py
```diff
@@ -441,6 +441,42 @@ def get_tag_or_default(alignment: pysam.AlignedSegment, tag_key: str, default: O
         return default
 
 
+def sort_by_tags_and_queryname(
+        records: Iterable[pysam.AlignedSegment],
+        tags: Iterable[str] = None) -> Iterable[pysam.AlignedSegment]:
```
ambrosejcarr: I think this type is probably `Optional[Iterable[str]]` since it's assigned `None`.

src/sctools/bam.py
```diff
@@ -441,6 +441,42 @@ def get_tag_or_default(alignment: pysam.AlignedSegment, tag_key: str, default: O
...
+        tags = []
+    comparable_records = []
+    for r in records:
+        cr = TagSortableRecord.from_aligned_segment(r, tags)
+        comparable_records.append(cr)
+    sorted_records = sorted(comparable_records)
+    return TagSortableRecord.get_aligned_segments(sorted_records)
+
+
+def read_bam(input_path: str) -> (pysam.AlignmentFile.header, List[pysam.AlignedSegment]):
```
ambrosejcarr: Interesting, does mypy validate explicit tuples? My experience is that this should be `Tuple[pysam.AlignmentFile.header, List[pysam.AlignedSegment]]`

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
...
+    def __lt__(self, other: 'TagSortableRecord'):
+        self.__verify_tag_keys_match(other)
+        for i in range(len(self.tag_keys)):
+            if self.tag_values[i] < other.tag_values[i]:
+                return True
+            elif self.tag_values[i] > other.tag_values[i]:
+                return False
+        return self.query_name < other.query_name
+
+    def __eq__(self, other: 'TagSortableRecord'):
```
ambrosejcarr: I would add a #TODO here that this block could probably contain more error checking. It would take some careful thought, but there are some weird edge cases implied by the sam format that this could catch.

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
...
+
+    def __repr__(self):
+        format_str = 'TagSortableRecord(tags: {0}, tag_values: {1}, query_name: {2}'
+        return format_str.format(self.tag_keys, self.tag_values, self.query_name)
 
-    @property
-    def key_generator(self) -> Callable[[pysam.AlignedSegment], Tuple[str, str, str, str]]:
-        return self._get_sort_key
+    @staticmethod
+    def get_aligned_segments(comparable_records: Iterable['TagSortableRecord']) -> Iterable[pysam.AlignedSegment]:
```
ambrosejcarr: I know I said "read it all into memory" but I would delegate that read to the latest possible point so that we're left with more flexible code for later. Ignore this comment if it would be a lot of work, but I'd make this return `Iterator[AlignedSegment]` and save the read as last as possible.

src/sctools/bam.py
```diff
@@ -441,6 +441,42 @@ def get_tag_or_default(alignment: pysam.AlignedSegment, tag_key: str, default: O
         return default
 
 
+def sort_by_tags_and_queryname(
```
ambrosejcarr: Since this is the key object in this module, I'd put it at the bottom.

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
         return 'query_name'
 
 
-class CellMoleculeGeneQueryNameSortOrder(AlignmentSortOrder):
-    """Hierarchical alignment record sort order (cell barcode >= molecule barcode >= gene name >= query name)."""
+@functools.total_ordering
+class TagSortableRecord(object):
```
ambrosejcarr: mypy says that Init should return None. `__init__() -> None:`

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
...
+        if tag_keys is None:
+            tag_keys = []
+        tag_values = []
+        for key in tag_keys:
+            tag_value = get_tag_or_default(record, key, '')
+            tag_values.append(tag_value)
+        query_name = record.query_name
+        return cls(tag_keys, tag_values, query_name, record)
+
+    def __lt__(self, other: 'TagSortableRecord'):
```
ambrosejcarr: returns a bool (same eq)

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
...
+        return self.query_name < other.query_name
+
+    def __eq__(self, other: 'TagSortableRecord'):
+        self.__verify_tag_keys_match(other)
+        for i in range(len(self.tag_values)):
+            if self.tag_values[i] != other.tag_values[i]:
+                return False
+        return self.query_name == other.query_name
+
+    def __verify_tag_keys_match(self, other):
```
ambrosejcarr: returns None

src/sctools/bam.py
```diff
@@ -464,29 +500,78 @@ def __repr__(self) -> str:
...
+        if self.tag_keys != other.tag_keys:
+            format_str = 'Cannot compare records using different tag lists: {0}, {1}'
+            raise ValueError(format_str.format(self.tag_keys, other.tag_keys))
+
+    def __str__(self):
```
ambrosejcarr: returns string (also repr, below).

src/sctools/bam.py
dshiga: You're right, it wants it the way you said. Will fix.

PR #47, opened 2018-08-29
Jx ss2 group qc

jishuxu:
add python scripts to parse and group SS2 QC outputs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup(
                 'MergeCountMatrices = sctools.platform:GenericPlatform.merge_count_matrices',
                 'TagSortBam = sctools.platform:GenericPlatform.tag_sort_bam',
                 'VerifyBamSort = sctools.platform:GenericPlatform.verify_bam_sort',
-                'GroupQCs = sctools.platform.GenericPlatform.group_qc_outputs'
+                'GroupQCs = sctools.platform:GenericPlatform.group_qc_outputs'
             ]
     },
     classifiers=CLASSIFIERS,
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,8 @@ setup(
         'sphinx_rtd_theme',
         'setuptools_scm',
         'scipy>=1.0.0',
+        'crimson>=0.3.0',
+
     ],
     entry_points={
             'console_scripts': [
@@ -45,7 +47,8 @@ setup(
                 'CreateCountMatrix = sctools.platform:GenericPlatform.bam_to_count_matrix',
                 'MergeCountMatrices = sctools.platform:GenericPlatform.merge_count_matrices',
                 'TagSortBam = sctools.platform:GenericPlatform.tag_sort_bam',
-                'VerifyBamSort = sctools.platform:GenericPlatform.verify_bam_sort'
+                'VerifyBamSort = sctools.platform:GenericPlatform.verify_bam_sort',
+                'GroupQCs = sctools.platform.GenericPlatform.group_qc_outputs'
             ]
     },
     classifiers=CLASSIFIERS,
--- a/src/sctools/__init__.py
+++ b/src/sctools/__init__.py
@@ -8,6 +8,7 @@ from . import reader
 from . import metrics
 from . import platform
 from . import consts
+from . import groups
 from pkg_resources import get_distribution, DistributionNotFound
 
 
--- a/src/sctools/groups.py
+++ b/src/sctools/groups.py
@@ -34,7 +34,12 @@ def AggregatePicardMetricsRow(filenames, output_name):
         # but only output PAIRED-READS/third line
         if class_name == "AlignmentSummaryMetrics":
             # only parse out pair reads
-            met = parsed['metrics']['contents'][2]
+            y = {}
+            x = parsed['metrics']['contents']
+            for m in x:
+                cat = m['CATEGORY']
+                y.update({k + '.' + cat: v for k, v in m.items() if k not in ['SAMPLE', 'LIBRARY', 'READ_GROUP', 'CATEGORY']})
+            met = y
         # sometimes(very rare), insertion metrics also return multiple lines
         # results to include TANDEM repeats. but we only output the first line.
         elif class_name == "InsertSizeMetrics":
@@ -49,7 +54,7 @@ def AggregatePicardMetricsRow(filenames, output_name):
             met = parsed['metrics']['contents']
         mets[cell_id].update({
                 k: met[k] for k in met if k not in
-                ['SAMPLE', 'LIBRARY', 'READ_GROUP']
+                ['SAMPLE', 'LIBRARY', 'READ_GROUP', 'CATEGORY']
                 })
         df = pd.DataFrame.from_dict(mets, orient='columns')
         df.insert(0, 'Class', class_name)
--- a/src/sctools/groups.py
+++ b/src/sctools/groups.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Created on Tue Aug 21 11:07:46 2018
+
+@author: jishuxu
+"""
+
+from crimson import picard
+import argparse
+import os
+import pandas as pd
+
+
+def AggregatePicardMetricsRow(filenames, output_name):
+    """
+    pipeline output picard QC metrics at sinle cell/sample level.
+    This functin is called to merge/aggregate QC metrics by metrics type and
+    then merge multiple QC measurement into single matrix file.
+    In this file, column is sample/cell and row is QC metrics
+    :param files: metric files from pipeline outputs
+    :param met_name: metrics name with workflow name and subworkflow name
+    as prefix.such as 'run_pipelines.RunStarPipeline.alignment_summary_metrics'
+    """
+    # initial output
+    mets = {}
+    d = pd.DataFrame()
+    for file_name in filenames:
+        cell_id = os.path.basename(file_name).split('_qc')[0]
+        mets[cell_id] = {}
+        parsed = picard.parse(file_name)
+        class_name = parsed['metrics']['class'].split('.')[2]
+        # Aignment metrics return multiple lines,
+        # but only output PAIRED-READS/third line
+        if class_name == "AlignmentSummaryMetrics":
+            # only parse out pair reads
+            met = parsed['metrics']['contents'][2]
+        # sometimes(very rare), insertion metrics also return multiple lines
+        # results to include TANDEM repeats. but we only output the first line.
+        elif class_name == "InsertSizeMetrics":
+            # if the elemnet counts is less than 21,
+            # it means insertion metrics returns multiple line results.
+            if len(parsed['metrics']['contents']) < 21:
+                met = parsed['metrics']['contents'][0]
+            else:
+                met = parsed['metrics']['contents']
+        else:
+            # other metrics(so far) only return one line results.
+            met = parsed['metrics']['contents']
+        mets[cell_id].update({
+                k: met[k] for k in met if k not in
+                ['SAMPLE', 'LIBRARY', 'READ_GROUP']
+                })
+        df = pd.DataFrame.from_dict(mets, orient='columns')
+        df.insert(0, 'Class', class_name)
+        d = d.append(df)
+    d.to_csv(output_name + '.csv')
+
+
+def AggregatePicardMetricsTable(filenames, output_name):
+    for file_name in filenames:
+        cell_id = os.path.basename(file_name).split('_qc')[0]
+        class_name = os.path.basename(file_name).split('.')[1]
+        parsed = picard.parse(file_name)
+        dat = pd.DataFrame.from_dict(parsed['metrics']['contents'])
+        dat.insert(0, 'Sample', cell_id)
+        dat.to_csv(output_name + "_"+class_name+'.csv', index=False)
+
+
+def AggregateQCMetrics(filenames, output_name):
+    df = pd.DataFrame()
+    for file_name in filenames:
+        dat = pd.read_csv(file_name)
+        df = df.append(dat, ignore_index=True)
+    df.rename(columns={'Unnamed: 0': 'Metrics'}, inplace=True)
+    df.to_csv(output_name + '.csv', index=False)
+
+
+def ParseHISATStats(filenames, output_name):
+    mets = {}
+    for file_name in filenames:
+        if '_qc' in file_name:
+            cell_id = os.path.basename(file_name).split('_qc')[0]
+            tag = "HISAT2G"
+        elif '_rsem' in file_name:
+            cell_id = os.path.basename(file_name).split('_rsem')[0]
+            tag = "HISAT2T"
+        with open(file_name) as f:
+            dat = f.readlines()
+            d = [x.strip().split(':') for x in dat]
+            d.pop(0)
+            mets[cell_id] = {x[0]: x[1].strip().split(' ')[0] for x in d}
+    df = pd.DataFrame.from_dict(mets, orient='columns')
+    df.insert(0, "Class", tag)
+    df.to_csv(output_name + '.csv')
+
+
+def ParseRSENStats(filenames, output_name):
+    mets = {}
+    for file_name in filenames:
+        cell_id = os.path.basename(file_name).split('_rsem')[0]
+        i = 0
+        with open(file_name) as f:
+            while i < 3:
+                if i == 0:
+                    [N0, N1, N2, N_tot] = f.readline().strip().split(" ")
+                elif i == 1:
+                    [nUnique, nMulti, nUncertain] = \
+                        f.readline().strip().split(" ")
+                elif i == 2:
+                    [nHits, read_type] = f.readline().strip().split(" ")
+                i = i+1
+        mets[cell_id] = {
+                "unalignable reads": N0,
+                "alignable reads": N1,
+                "filtered reads": N2,
+                "total reads": N_tot,
+                "unique aligned": nUnique,
+                "multiple mapped": nMulti,
+                "total alignments": nHits,
+                "strand": read_type,
+                "uncertain reads": nUncertain
+        }
+    df = pd.DataFrame.from_dict(mets, orient='columns')
+    df.insert(0, "Class", "RSEM")
+    df.to_csv(output_name + '.csv')
--- a/src/sctools/platform.py
+++ b/src/sctools/platform.py
@@ -421,7 +421,7 @@ class GenericPlatform:
 
         return 0
     
-    def group_qc_outputs() :
+    def group_qc_outputs():
         """Commandline entrypoint for parsing picard metrics files, hisat2 and rsem statistics log files.
         Parameters
         ----------
@@ -440,7 +440,6 @@ class GenericPlatform:
                 nargs='+',
                 required=True,
                 help="a list of files to be parsed out.")
-        
         parser.add_argument(
                 "-o",
                 "--output_name",
@@ -455,18 +454,18 @@ class GenericPlatform:
                 help="a list of string to represent metrics types,such Picard, PicardTable, HISAT2,RSEM, Core")
         args = parser.parse_args()
         if args.mtype == "Picard":
-            AggregatePicardMetricsRow(args.file_names, args.output_name)
+            groups.AggregatePicardMetricsRow(args.file_names, args.output_name)
         elif args.mtype == "PicardTable":
-            AggregatePicardMetricsTable(args.file_names, args.output_name)
+            groups.AggregatePicardMetricsTable(args.file_names, args.output_name)
         elif args.mtype == "Core":
-            AggregateQCMetrics(args.file_names, args.output_name)
+            groups.AggregateQCMetrics(args.file_names, args.output_name)
         elif args.mtype == "HISAT2":
-            ParseHISATStats(args.file_names, args.output_name)
+            groups.ParseHISATStats(args.file_names, args.output_name)
         elif args.mtype == "RSEM":
-            ParseRSENStats(args.file_names, args.output_name)
+            groups.ParseRSENStats(args.file_names, args.output_name)
         else:
             return 0
-    
+
 class TenXV2(GenericPlatform):
     """Command Line Interface for 10x Genomics v2 RNA-sequencing programs
 
--- a/src/sctools/platform.py
+++ b/src/sctools/platform.py
@@ -22,7 +22,7 @@ from typing import Iterable, List, Dict, Optional, Sequence
 from itertools import chain
 
 import pysam
-from sctools import fastq, bam, metrics, count, consts, gtf
+from sctools import fastq, bam, metrics, count, consts, gtf, groups
 
 
 class GenericPlatform:
@@ -48,6 +48,8 @@ class GenericPlatform:
         construct a compressed sparse row count file from a tagged, aligned bam file
     merge_count_matrices()
         merge multiple csr-format count matrices into a single csr matrix
+    group_qc_outputs()
+        aggregate Picard, HISAT2 and RSME QC statisitics
     """
 
     @classmethod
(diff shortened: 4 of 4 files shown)


src/sctools/platform.py
```diff
@@ -418,7 +420,51 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
...
+                help="a list of files to be parsed out.")
+        parser.add_argument(
+                "-o",
+                "--output_name",
+                dest="output_name",
+                required=True,
+                help="The output file name")
+        parser.add_argument(
+                "-t",
+                "--metrics-type",
```
samanehsan: Please use `metrics_type` for consistency with the other arguments

src/sctools/platform.py
```diff
@@ -418,7 +420,51 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
...
+            groups.AggregatePicardMetricsRow(args.file_names, args.output_name)
+        elif args.mtype == "PicardTable":
+            groups.AggregatePicardMetricsTable(args.file_names, args.output_name)
+        elif args.mtype == "Core":
+            groups.AggregateQCMetrics(args.file_names, args.output_name)
+        elif args.mtype == "HISAT2":
+            groups.ParseHISATStats(args.file_names, args.output_name)
+        elif args.mtype == "RSEM":
+            groups.ParseRSENStats(args.file_names, args.output_name)
+        else:
```
samanehsan: This `else` is unnecessary here, you can just `return 0`

src/sctools/platform.py
samanehsan: ^ @jishuxu

src/sctools/groups.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def aggregate_picard_metrics_by_row(file_names, output_name):
+    """
+    pipeline output picard QC metrics at sinle cell/sample level.
+    This functin is called to merge/aggregate QC metrics by metrics type and
+    then merge multiple QC measurement into single matrix file.
+    In this file, column is sample/cell and row is QC metrics
+    :param files: metric files from pipeline outputs
+    :param met_name: metrics name with workflow name and subworkflow name
```
samanehsan: Can you please change these docstring parameters to `file_names` and `output` name to reflect the current function parameters?

src/sctools/groups.py
```diff
@@ -0,0 +1,127 @@
...
+
+from crimson import picard
+import argparse
+import os
+import pandas as pd
+
+
+def aggregate_picard_metrics_by_row(file_names, output_name):
+    """
+    pipeline output picard QC metrics at sinle cell/sample level.
```
samanehsan: Tiny nit: typo in sinle --> single

src/sctools/groups.py
```diff
@@ -0,0 +1,127 @@
...
+    This functin is called to merge/aggregate QC metrics by metrics type and
```
samanehsan: Another small typo: functin --> function

src/sctools/groups.py
```diff
@@ -0,0 +1,127 @@
...
+                k: rows[k] for k in rows if k not in
+                ['SAMPLE', 'LIBRARY', 'READ_GROUP', 'CATEGORY']
+                })
+        df = pd.DataFrame.from_dict(metrics, orient='columns')
+        df.insert(0, 'Class', class_name)
+        d = d.append(df)
+    d.to_csv(output_name + '.csv')
+
+
+def aggregate_picard_metrics_by_table(file_names, output_name):
```
samanehsan: Please add a brief docstring to these other functions as well!

setup.py
```diff
@@ -33,6 +33,8 @@
         'sphinx_rtd_theme',
         'setuptools_scm',
         'scipy>=1.0.0',
+        'crimson>=0.3.0',
```
samanehsan: Please add `crimson>=0.3.0` to [requirements.txt](https://github.com/HumanCellAtlas/sctools/blob/master/requirements.txt).

This should also fix the failing Circle CI tests because they install requirements from the requirements.txt file: https://github.com/HumanCellAtlas/sctools/blob/master/.circleci/config.yml#L34

src/sctools/groups.py
samanehsan: Thanks for adding that in @jishuxu! Can you add docstrings to the rest of the functions as well?

src/sctools/groups.py
```diff
@@ -0,0 +1,127 @@
...
+        elif '_rsem' in file_name:
+            cell_id = os.path.basename(file_name).split('_rsem')[0]
+            tag = "HISAT2T"
+        with open(file_name) as f:
+            dat = f.readlines()
+            d = [x.strip().split(':') for x in dat]
+            d.pop(0)
+            metrics[cell_id] = {x[0]: x[1].strip().split(' ')[0] for x in d}
+    df = pd.DataFrame.from_dict(metrics, orient='columns')
+    df.insert(0, "Class", tag)
```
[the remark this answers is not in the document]
samanehsan: @jishuxu, is it possible to have both `_qc` and `_rsem` files in `file_names`? What should the tag be in that situation? Currently `tag` would be the tag of the last file in `file_names`.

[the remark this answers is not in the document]
jishuxu: no, we add _qc and _rsem to differentiate two tracks in pipeline, one track is _qc and another is _rsem. so it is impossible to have both suffix in filename.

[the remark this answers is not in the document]
samanehsan: Okay, thanks for the explanation!

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    then merge multiple QC measurement into single matrix file.
+    In this file, column is sample/cell and row is QC metrics
+    :param file_names: list of files generated by Picard
+    :param output_name: output name without suffix'
+    """
+    # initial output
+    metrics = {}
+    d = pd.DataFrame()
+    for file_name in file_names:
+        cell_id = os.path.basename(file_name).split('_qc')[0]
```
ambrosejcarr: This feels dependent upon the naming scheme. Can you list the required name format for this tool to work? And, if not standard, which tool's output it should process?

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+        if class_name == "AlignmentSummaryMetrics":
+            # parse out PE, R1 and R2
+            rows = {}
+            for m in contents:
+                cat = m['CATEGORY']
+                rows.update({k + '.' + cat: v for k, v in m.items() if k not in ['SAMPLE', 'LIBRARY', 'READ_GROUP', 'CATEGORY']})
+        # sometimes(very rare), insertion metrics also return multiple lines
+        # results to include TANDEM repeats. but we only output the first line.
+        elif class_name == "InsertSizeMetrics":
+            # if the elemnet counts is less than 21,
```
ambrosejcarr: element

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+
+from crimson import picard
+import argparse
+import os
+import pandas as pd
+
+
+def aggregate_picard_metrics_by_row(file_names, output_name):
+    """
+    pipeline output picard QC metrics at single cell/sample level.
```
ambrosejcarr:
```diff
+ pipeline outputs
- pipeline output
```

Also line 62 below.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+        d = d.append(df)
+    d.to_csv(output_name + '.csv')
+
+
+def aggregate_picard_metrics_by_table(file_names, output_name):
+    """
+    pipeline output picard QC metrics at single cell/sample level.
+    This function is called to merge/aggregate QC metrics by metrics type and
+    then merge multiple QC measurement into single matrix file.
+    In this file, column is sample/cell and row is QC metrics
```
ambrosejcarr: Would prefer this be transposed. What was the motivation for this orientation?

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    :param file_names: list of files generated by Picard, metrics are in table.
+    :param output_name: output name without suffix'
```
ambrosejcarr: trailing `'`

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+   """
+    for file_name in file_names:
```
ambrosejcarr: Same as above, would like examples of what file names should look like.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+        cell_id = os.path.basename(file_name).split('_qc')[0]
+        class_name = os.path.basename(file_name).split('.')[1]
+        parsed = picard.parse(file_name)
+        dat = pd.DataFrame.from_dict(parsed['metrics']['contents'])
+        dat.insert(0, 'Sample', cell_id)
+        dat.to_csv(output_name + "_" + class_name + '.csv', index=False)
```
ambrosejcarr: Multiple separators used: `.` used above, `_` used here. Is there a reason these are different?

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def aggregate_qc_metrics(file_names, output_name):
+    """
+    Append several metrics into one file.
```
ambrosejcarr: I think it is worth adding that this writes a file.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+        df.insert(0, 'Class', class_name)
+        d = d.append(df)
+    d.to_csv(output_name + '.csv')
+
+
+def aggregate_picard_metrics_by_table(file_names, output_name):
+    """
+    pipeline output picard QC metrics at single cell/sample level.
+    This function is called to merge/aggregate QC metrics by metrics type and
+    then merge multiple QC measurement into single matrix file.
```
ambrosejcarr: ... which is written to `.csv`.

Same on line 16 above.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    for file_name in file_names:
+        dat = pd.read_csv(file_name)
+        df = df.append(dat, ignore_index=True)
+    df.rename(columns={'Unnamed: 0': 'Metrics'}, inplace=True)
+    df.to_csv(output_name + '.csv', index=False)
+
+
+def parse_hisat2_log(file_names, output_name):
+    """
+    to parse the hisat2 log file, which include alignment summary
```
ambrosejcarr: Please make this a complete sentence.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    :param file_names: hisat2 log files
+    :param output_name: output file name without suffix
+    """
+    metrics = {}
+    tag = "NONE"
+    for file_name in file_names:
```
ambrosejcarr: Talk about the logic here -- what do these `_qc` an `_rsem` naming conventions imply? Why are they parsed differently?

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+        if '_qc' in file_name:
+            cell_id = os.path.basename(file_name).split('_qc')[0]
+            tag = "HISAT2G"
+        elif '_rsem' in file_name:
+            cell_id = os.path.basename(file_name).split('_rsem')[0]
+            tag = "HISAT2T"
+        with open(file_name) as f:
+            dat = f.readlines()
+            d = [x.strip().split(':') for x in dat]
+            d.pop(0)
```
ambrosejcarr: Please add some comments about these magic numbers. What's going on with the pop, splitting?

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+            metrics[cell_id] = {x[0]: x[1].strip().split(' ')[0] for x in d}
+    df = pd.DataFrame.from_dict(metrics, orient='columns')
+    df.insert(0, "Class", tag)
+    df.to_csv(output_name + '.csv')
+
+
+def parse_rsem_cnt(file_names, output_name):
+    """
+    to parse the rsem cnt file, which include alignment summary
```
ambrosejcarr:
```diff
+ which includes
- which include
```

src/sctools/platform.py
```diff
@@ -418,8 +420,54 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
...
+        elif args.metrics_type == "Core":
+            groups.aggregate_qc_metrics(args.file_names, args.output_name)
+        elif args.metrics_type == "HISAT2":
+            groups.parse_hisat2_log(args.file_names, args.output_name)
+        elif args.metrics_type == "RSEM":
+            groups.parse_rsem_cnt(args.file_names, args.output_name)
+        return 0
 
-
+
```
ambrosejcarr: extra white space

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    :param file_names: rsem cnt files
+    :param output_name: output file name without suffix
+    """
+    metrics = {}
```
ambrosejcarr: more or less same comments here about explaining a bit more about what is going on.

src/sctools/groups.py
```diff
@@ -0,0 +1,151 @@
...
+    for file_name in file_names:
+        cell_id = os.path.basename(file_name).split('_rsem')[0]
+        i = 0
+        with open(file_name) as f:
+            while i < 3:
+                if i == 0:
+                    [N0, N1, N2, N_tot] = f.readline().strip().split(" ")
+                elif i == 1:
+                    [nUnique, nMulti, nUncertain] = \
```
ambrosejcarr: `n_unique` in python.

src/sctools/groups.py
jishuxu: the output of this way would be "a long skinny" table(not vector). The first 2 or 3 columns are meta data related to Picard metrics and the last column is the actual measurement. I think it is more traditional way to output Picard metrics.

src/sctools/groups.py
jishuxu: Functional-wise, it does the same thing as another function. I added a few words to describe more details

src/sctools/groups.py
jishuxu: it will pop out the first row of each section.

src/sctools/groups.py
ambrosejcarr: Ok, my motivation is that consistency of our object model is important. In our matrix discussions, we came to the conclusion that features/variables as columns and observations as rows was what we'd like to go with. 

Transposing this matrix would be consistent with that. I don't find the long vs. tall discussion compelling because a small SS2 experiment also outputs a long matrix (~1000 cells x ~40,000-60,000 features).

jishuxu: we extract two different metrics from Picard outputs, one is the row-based and another is the table-based. For the row based, it is certainly we can transpose them into any ways. For the table-based, such as base distribution, gc bias distribution, error distribution.... it probably is not necessary to transpose them at all.

src/sctools/platform.py
```diff
@@ -418,6 +420,52 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
         count_matrix.save(args.output_stem)
 
         return 0
+    
+    def group_qc_outputs():
```
ambrosejcarr: I would make this `group_picard_qc_metrics` -- it won't work for other metrics in this package.

src/sctools/groups.py
```diff
@@ -0,0 +1,160 @@
...
+
+from crimson import picard
+import argparse
+import os
+import pandas as pd
+
+
+def aggregate_picard_metrics_by_row(file_names, output_name):
+    """
+    pipeline outputs picard QC metrics at single cell/sample level.
```
ambrosejcarr: The rest of SCTools uses numpy docblocks, can you convert these?

src/sctools/platform.py
```diff
@@ -418,6 +420,52 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
...
+        """Commandline entrypoint for parsing picard metrics files, hisat2 and rsem statistics log files.
+        Parameters
```
ambrosejcarr: I think the right numpydoc block for this is: 

```python
"""
Command line entry-point for parsing Picard, HISAT2, and RSEM statistics log files.

        Returns
        ----------
        int:
            return 0 if the program completes successfully.
"""
```

src/sctools/platform.py
```diff
@@ -418,6 +420,52 @@ def merge_count_matrices(cls, args: Iterable[str]=None) -> int:
...
+                "--output_name",
+                dest="output_name",
+                required=True,
+                help="The output file name")
+        parser.add_argument(
+                "-t",
+                "--metrics_type",
+                dest="metrics_type",
+                required=True,
+                help="a list of string to represent metrics types,such Picard, PicardTable, HISAT2,RSEM, Core")
```
ambrosejcarr: You can use the `options` parameter here to define valid values.

src/sctools/groups.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def parse_rsem_cnt(file_names, output_name):
+    """
+    to parse the rsem cnt file to return a set of statistics, such as 
+    total alignment used, total multiple mapped alignment, total un-aligned reads
+    :param file_names: rsem cnt files
+    :param output_name: output file name without suffix
+    """
+    metrics = {}
```
ambrosejcarr: Need an example filename that will work for this parser.

src/sctools/__init__.py
```diff
@@ -8,6 +8,7 @@
 from . import metrics
 from . import platform
 from . import consts
+from . import groups
```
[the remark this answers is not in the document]
jishuxu: it is not only picards, there are more than picard

src/sctools/platform.py
jishuxu: add `choices`

src/sctools/platform.py
jishuxu: changed

src/sctools/__init__.py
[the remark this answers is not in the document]
ambrosejcarr: Yep, that's right. Not `picard`, but `groups` is not descriptive enough, and all the modules _are_ metrics related, so I do think it belongs in there. 

`metrics.merge`, maybe?

src/sctools/groups.py
```diff
@@ -4,19 +4,21 @@
...
 
-def aggregate_picard_metrics_by_row(file_names, output_name):
-    """
-    pipeline outputs picard QC metrics at single cell/sample level.
-    This function is called to merge/aggregate QC metrics by metrics type and
-    then merge multiple QC measurement into single matrix file.
-    In this file, column is sample/cell and row is QC metrics
-    :param file_names: list of files generated by Picard
-    :param output_name: output name without suffix'
+def write_aggregated_picard_metrics_by_row(file_names, output_name):
```
ambrosejcarr: These are close, but you can pattern-match on the existing docs: https://github.com/HumanCellAtlas/sctools/blob/master/src/sctools/barcode.py#L37

src/sctools/platform.py
ambrosejcarr: There are some whitespace inconsistencies in the help string. Also, this isn't a list, is it? It's a single string?

src/sctools/groups.py
jishuxu: added in docstring

src/sctools/groups.py
jishuxu: changed

src/sctools/groups.py
jishuxu: changed

src/sctools/groups.py
jishuxu: changed

src/sctools/groups.py
jishuxu: changed

src/sctools/platform.py
jishuxu: changed

src/sctools/groups.py
```diff
@@ -54,15 +54,15 @@ def aggregate_picard_metrics_by_row(file_names, output_name):
         df = pd.DataFrame.from_dict(metrics, orient='columns')
         df.insert(0, 'Class', class_name)
         d = d.append(df)
-    d.to_csv(output_name + '.csv')
+    d_T = d.T
+    d_T.to_csv(output_name + '.csv')
 
 
 def aggregate_picard_metrics_by_table(file_names, output_name):
```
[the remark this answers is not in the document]
jishuxu: changed

src/sctools/groups.py
```diff
@@ -0,0 +1,160 @@
...
+
+"""
+
+from crimson import picard
+import argparse
+import os
+import pandas as pd
+
+
+def aggregate_picard_metrics_by_row(file_names, output_name):
```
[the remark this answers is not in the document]
jishuxu: changed

src/sctools/groups.py
jishuxu: changed

src/sctools/groups.py
```diff
@@ -66,7 +66,8 @@ def write_aggregated_picard_metrics_by_table(file_names, output_name):
     ----------
     args:
         file_names: array of files.the basename of inputs should be formated as 'samplename_qc'
-        output_name: prefix of output file name. the basename of outputs include the Pciard
+        output_name: prefix of output file name. the basename of outputs
+        includes the Pciard
```
kbergin: Typo in `Picard`

kbergin:
@chrisfarnham

setup.py
```diff
@@ -45,7 +47,8 @@
                 'CreateCountMatrix = sctools.platform:GenericPlatform.bam_to_count_matrix',
                 'MergeCountMatrices = sctools.platform:GenericPlatform.merge_count_matrices',
                 'TagSortBam = sctools.platform:GenericPlatform.tag_sort_bam',
-                'VerifyBamSort = sctools.platform:GenericPlatform.verify_bam_sort'
+                'VerifyBamSort = sctools.platform:GenericPlatform.verify_bam_sort',
+                'GroupQCs = sctools.platform:GenericPlatform.group_qc_outputs'
```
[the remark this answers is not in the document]
ambrosejcarr: Maybe GroupSS2QCs?

src/sctools/groups.py
ambrosejcarr: Got it! with the added docstring, I think I understand now. 

This would be clearer as 
```python3
# trim _qc suffix
basename = os.path.basename(file_name)
if basename.endswith('_qc'):
    cell_id = basename[:-3]
else:
    raise ValueError(
        "QC metric file not properly formatted. Please ensure that file names are formatted as 
        '<cell_id>_qc'."
    )
```

src/sctools/groups.py
ambrosejcarr: See above suggestion with respect to making the code clearer. In addition, 

```
        class_name = os.path.basename(file_name).split('.')[1]
```
this makes it seem like there are more requirements than just `samplename_qc`. I still don't understand this, because `cell_id` somehow contains `class_name` by the logic here: 

```
        cell_id = os.path.basename(file_name).split('_qc')[0]
        class_name = os.path.basename(file_name).split('.')[1]
```

src/sctools/groups.py
ambrosejcarr: Then I think line 39 above should be `_` by this logic? https://github.com/HumanCellAtlas/sctools/pull/47/files#diff-575284da92642b59c81fffaa1a3936f0R39

src/sctools/groups.py
ambrosejcarr: Cool. Why are you doing this? Is it a header that you don't need? (put it into the code comment, please)

src/sctools/groups.py
```diff
@@ -0,0 +1,179 @@
...
+    df_T = df.T
+    df_T.to_csv(output_name + '.csv')
+
+
+def parse_rsem_cnt(file_names, output_name):
+    """Command line entrypoint parse, aggreagete and write RSEM cnt
+    Parameters
+    ----------
+    args:
+        file_names: array of RSEM cnt files. The basename of inputs should be '_rsem'
```
ambrosejcarr: basename is the full file name less the directory structure -- I think this refers to the suffix. However, I'd love to know what the base name needs to be, here.

src/sctools/groups.py
jishuxu: the input name as "SRRXXX_qc.bait_bias_summary_metrics.txt" and  output is "SRRXXXX_qc_bait_bias_summary_metrics.csv"

setup.py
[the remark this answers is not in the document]
jishuxu: I also prefer not to do this right, because this change will affect on skylab and we have to update the pipeline as well.

samanehsan:
@ambrosejcarr since this PR is required for a few other important changes to go in, I'll create a ticket to document the remaining changes you suggested and we can include them in a follow-up PR.

PR #51, opened 2018-10-12
Make explicit that sctools does not support python 3.5

ambrosejcarr:
We don't support this. Remove lines from setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,9 +5,7 @@ CLASSIFIERS = [
     "Natural Language :: English",
     "License :: OSI Approved :: BSD License",
     "Operating System :: OS Independent",
-    "Programming Language :: Python :: 3.5",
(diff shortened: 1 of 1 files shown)


rexwangcc:
I don't know where I should put this comment, so let me comment on this PR... The docs build has been failing due to an outdated setuptools version. We need to either make sure readthedocs executes `pip install -U setuptools` or include the latest version of setuptools in requirements?

ambrosejcarr:
> include the latest version of setuptools in requirements?

I'd give that a shot -- `readthedocs` has some defaults pre-installed that don't make any sense (e.g. old versions of `numpy`).

ambrosejcarr:
^ This might also fix #52

rexwangcc:
Merging this now

PR #56, opened 2019-01-02
Handle KeyError caused by missing "CB" tags when calculating cell metrics

samanehsan:
### Purpose
Records with no "CB" tag will fail in the `CellMetrics.parse_extra_fields` method in the latest version (v0.3.1) of sctools. Due to this issue, this version cannot be used by the `CalculateCellMetrics` task of the Optimus pipeline. 

Fixes https://github.com/HumanCellAtlas/secondary-analysis/issues/455

Zenhub ticket: https://app.zenhub.com/workspaces/dcp-backlogs-5ac7bcf9465cb172b77760d9/issues/humancellatlas/secondary-analysis/455

---
### Changes
- Check that the record has a "CB" tag before checking that it is equivalent to the "CR" tag to avoid getting a KeyError. 
- Add test to verify `perfect_cell_barcodes`

---
### Review Instructions
- No instructions.

---
### PR Checklist
- [X] This PR added or updated tests.
- [X] This PR updated docstrings or documentation.
- [X] This PR applied Python style guidelines, specifically followed [NumPy style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) for this repo.
- [X] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions
- Should the number of records without a "CB" tag be stored as another variable? 
- The test data for records with missing CB tags was pulled from an Optimus run using `pbmc8k` data -- is there something else we could/should use instead?


src/sctools/metrics/aggregator.py
```diff
@@ -455,8 +455,12 @@ def parse_extra_fields(self, tags: Sequence[str], record: pysam.AlignedSegment)
             self._quality_above_threshold(
                 30, self._quality_string_to_numeric(record.get_tag(consts.QUALITY_CELL_BARCODE_TAG_KEY))))
 
-        self.perfect_cell_barcodes += (
-            record.get_tag(consts.RAW_CELL_BARCODE_TAG_KEY) == record.get_tag(consts.CELL_BARCODE_TAG_KEY))
+        try:
```
ambrosejcarr: Minor suggestion: break up line 459 into two lines, only cover the `get_tag` call that should trigger the key error with the `try: except` to ensure we don't also catch all the reads in case we receive data whose tags don't match our other constant for some reason.

src/sctools/test/test_metrics.py
```diff
@@ -33,10 +33,12 @@
 # set the input files
 _gene_sorted_bam = _data_dir + '/small-gene-sorted.bam'
 _cell_sorted_bam = _data_dir + '/small-cell-sorted.bam'
+_cell_sorted_bam_missing_cell_barcodes = _data_dir + '/cell-sorted-missing-cb.bam'
```
ambrosejcarr: would use `os.path.join()` instead of adding the slash for a bit more generality. Also line 41 below.

src/sctools/test/test_metrics.py
```diff
@@ -48,11 +50,16 @@
...
+cell_gatherer_missing_cbs = GatherCellMetrics(_cell_sorted_bam_missing_cell_barcodes, _cell_metric_output_file_missing_cell_barcodes)
+cell_gatherer_missing_cbs.extract_metrics()
+_cell_metrics_missing_cbs = pd.read_csv(_cell_metric_output_file_missing_cell_barcodes, index_col=0)
+
 
 def test_calculate_cell_metrics_cli():
     """test the sctools cell metrics CLI invocation"""
     return_call = TenXV2.calculate_cell_metrics(
-        args=['-i', _cell_sorted_bam, '-o', _test_dir + '/gene_metrics.csv'])
+        args=['-i', _cell_sorted_bam, '-o', _test_dir + '/cell_metrics.csv'])
```
ambrosejcarr: also `os.path.join`

src/sctools/test/test_metrics.py
```diff
@@ -144,12 +151,22 @@ def test_metrics_highest_read_count(metrics, expected_value):
     (_gene_metrics, 300),  # todo this is 100%, we should mangle a few in the testing data
     (_cell_metrics, 655),
 ])
-def test_metrics_number_perfect_barcodes(metrics, expected_value):
-    """Test that each metric correctly identifies the number of perfect barcodes where CB == CR"""
+def test_metrics_number_perfect_molecule_barcodes(metrics, expected_value):
+    """Test that each metric correctly identifies the number of perfect molecule barcodes where UB == UR"""
```
ambrosejcarr: 👍 good catch.

samanehsan:
Thanks for the review @ambrosejcarr! I started breaking out the cell barcode comparison so that `record.get_tag(consts.CELL_BARCODE_TAG_KEY)` was only in the try/except but then I realized I could use the `record.has_tag` function instead of relying on try/except to check for that tag. Let me know what you think!

PR #57, opened 2019-02-19
Bc generalize attach barcodes

benjamincarlin:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_

---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

- No changes.
---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- No instructions.

---
### PR Checklist
_Please ensure the following when opening a PR:_

- [ ] This PR added or updated tests.
- [x] This PR updated docstrings or documentation.
- [x] This PR applied Python style guidelines, specifically followed [NumPy style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) for this repo.
- [x] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions
_Please append follow-up discussions and issues during the review process below:_

Possible Tests:

- test the possible barcodes combinations of cell barcode, molecule barcode and sample barcode:
a test with only a cell barcode, a test with cell and sample barcodes, a test with cell and molecule barcodes, a test with cell, molecule and sample barcodes, a test with only a sample barcode, a test with a sample and molecule barcodes and a test with only molecule barcodes 

- test the possible barcode combinations that have a cell barcode with and without a whitelist

- test invalid start positions for cell, molecule and sample barcodes

- test invalid lengths for cell, molecule and sample barcodes

- test that cell and molecule barcodes do not overlap

- test that barcode start positions must be specified with barcode lengths

- test that an index fastq must be provided if a sample barcode start position and length are given

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ setup(
     ],
     entry_points={
             'console_scripts': [
-                'AttachBarcodes = sctools.platform:Attach.attach_barcodes',
+                'AttachBarcodes = sctools.platform:BarcodePlatform.attach_barcodes',
                 'Attach10xBarcodes = sctools.platform:TenXV2.attach_barcodes',
                 'SplitBam = sctools.platform:GenericPlatform.split_bam',
                 'CalculateGeneMetrics = sctools.platform:GenericPlatform.calculate_gene_metrics',
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
     ],
     entry_points={
             'console_scripts': [
+                'AttachBarcodes = sctools.platform:Attach.attach_barcodes',
                 'Attach10xBarcodes = sctools.platform:TenXV2.attach_barcodes',
                 'SplitBam = sctools.platform:GenericPlatform.split_bam',
                 'CalculateGeneMetrics = sctools.platform:GenericPlatform.calculate_gene_metrics',
--- a/src/sctools/platform.py
+++ b/src/sctools/platform.py
@@ -477,10 +477,8 @@ class GenericPlatform:
 
 class TenXV2(GenericPlatform):
     """Command Line Interface for 10x Genomics v2 RNA-sequencing programs
-
     This class defines several methods that are created as CLI tools when sctools is installed
     (see setup.py)
-
     Attributes
     ----------
     cell_barcode : fastq.EmbeddedBarcode
@@ -492,13 +490,11 @@ class TenXV2(GenericPlatform):
     sample_barcode : fastq.EmbeddedBarcode
         A data class that defines the start and end position of the sample barcode and the tags
         to assign the sequence and quality of the sample barcode
-
     Methods
     -------
     attach_barcodes()
         Attach barcodes from the forward (r1) and optionally index (i1) fastq files to the reverse
         (r2) bam file
-
     """
     # 10x contains three barcodes embedded within sequencing reads. The below objects define the
     # start and end points of those barcodes relative to the start of the sequence, and the
@@ -526,10 +522,8 @@ class TenXV2(GenericPlatform):
             output_bamfile_name: str,
             tag_generators: Iterable[fastq.EmbeddedBarcodeGenerator]) -> None:
         """Adds tags from fastq file(s) to a bam file.
-
         Attaches tags extracted from fastq files by `tag_generators`, attaches them to records from
         `input_bamfile_name`, and writes the result to `output_bamfile_name`
-
         Parameters
         ----------
         input_bamfile_name : str
@@ -538,20 +532,17 @@ class TenXV2(GenericPlatform):
             output bam
         tag_generators : Iterable[fastq.EmbeddedBarcodeGenerator]
             Iterable of generators that yield barcodes from fastq files
-
         """
         bam_tagger = bam.Tagger(input_bamfile_name)
         bam_tagger.tag(output_bamfile_name, tag_generators)
 
     @classmethod
     def _make_tag_generators(
-            cls, r1, i1=None, whitelist=None, no_cell_barcode=False, no_molecule_barcode=False) -> List[fastq.EmbeddedBarcodeGenerator]:
+            cls, r1, i1=None, whitelist=None) -> List[fastq.EmbeddedBarcodeGenerator]:
         """Create tag generators from fastq files.
-
         Tag generators are iterators that run over fastq records, they extract and yield all of the
         barcodes embedded in each fastq record. For 10x, this means extracting the cell, umi, and
         optionally, the sample barcode.
-
         Parameters
         ----------
         r1 : str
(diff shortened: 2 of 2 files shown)


src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+
+        Returns
+        -------
+        args : Iterable[str]
+            return arguments list if valid
+
+        """
+        # check that both the barcode length and position are given as arguments
+        # or that neither barcode length and position are given as arguments (XOR boolean logic)
+        cell_barcode_start_pos_exists = bool(args.cell_barcode_start_pos) or (args.cell_barcode_start_pos == 0)
```
ambrosejcarr: This functionality is repeated enough that I would write out a short function to replace it: 

```python
def _check_barcode_length_and_position(length: Union[int, str], position: int, name: str):
    """return True if both length and positions are provided, or neither are"""
    if isinstance(length, str):
        length = ast.literal_eval(length)
    if length =! bool(position): 
        raise argparse.ArgumentError(
            f"If either {name} barcode position or length are provided, both must be"
        )

```

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+    attach_barcodes()
+        Attach barcodes from the forward (r1) and optionally index (i1) fastq files to the reverse
+        (r2) bam file
+
+    """
+    cell_barcode = None
+    molecule_barcode = None
+    sample_barcode = None
+
+    @classmethod
```
ambrosejcarr: I suggest moving this below to directly above the two functions that call it.

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+
+        sample_barcode_start_pos_exists = bool(args.sample_barcode_start_pos) or (args.sample_barcode_start_pos == 0)
+        sample_barcode_length_exists = bool(args.sample_barcode_length)
+        if (sample_barcode_start_pos_exists != sample_barcode_length_exists):
+            argparse.ArgumentError('Invalid sample barocde position/length, both position and length must be provided by the user together')
+
+
+        # check that an index fastq is provided sample barcode length and position are given
+        if args.i1 is None and args.sample_barcode_length:
+            argparse.ArgumentError('An i7 index fastq file must be given to attach a sample barcode')
```
ambrosejcarr: I think this needs to be raised for the error to be thrown. 
```suggestion
            raise argparse.ArgumentError('An i7 index fastq file must be given to attach a sample barcode')
```

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+
+        return args
+
+    @classmethod
+    def _validate_barcode_start_pos(cls, given_value):
+        """Validates that the barcode start position is greater than 0
+
+        Parameters
+        ----------
+        given_value : int
```
ambrosejcarr: if given value is an `int`, then there is no need to call `int(given_value)` below. 

I suspect this should be `str` or `Union[int, str]`

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+        """
+        return cls._validate_barcode_input(int(given_value), 0)
+
+    @classmethod
+    def _validate_barcode_length(cls, given_value):
+        """Validates that the barcode length  is greater than 1
+
+        Parameters
+        ----------
+        given_value : int
```
ambrosejcarr: See above.

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+        Returns
+        -------
+        given_value : int
+            returns the length if it is valid
+
+        """
+        return cls._validate_barcode_input(int(given_value), 1)
+
+    @classmethod
+    def _tag_bamfile(cls,
```
ambrosejcarr: Cool -- this didn't need to change. I think this means it should be refactored to add it to the superclass (`GenericPlatform`) and removed from `TenXV2`

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+        """
+        bam_tagger = bam.Tagger(input_bamfile_name)
+        bam_tagger.tag(output_bamfile_name, tag_generators)
+
+    @classmethod
+    def _make_tag_generators(cls, r1, i1=None, whitelist=None) -> List[fastq.EmbeddedBarcodeGenerator]:
+        """Create tag generators from fastq files.
+
+        Tag generators are iterators that run over fastq records, they extract and yield all of the
+        barcodes embedded in each fastq record. For 10x, this means extracting the cell, umi, and
```
ambrosejcarr: Is this 10x? I suspect you need to modify this.

setup.py
```diff
@@ -37,6 +37,7 @@
     ],
     entry_points={
             'console_scripts': [
+                'AttachBarcodes = sctools.platform:BarcodePlatform.attach_barcodes',
```
ambrosejcarr: Can this be called something more descriptive? `AttachBarcodes` is so general that 10x users might think to use it for their method.

benjamincarlin: This is more general than the 10x, so users could use this as well. Is the a place where I could provide more explicit documentation explaining the difference between the two classes? (10xV2 has hardcoded values for the barcode start positions/lengths for that particular assay rather than requiring users to provide barcodes start positions/lengths for the ones they want)

src/sctools/platform.py
```diff
@@ -629,3 +629,304 @@ def attach_barcodes(cls, args=None):
...
+        return given_value
+
+    @classmethod
+    def _validate_barcode_args(cls, args):
+        """Validates that the barcode start position is greater than 0
+
+        Parameters
+        ----------
+        args : Iterable[str]
+            arguments list, The default value of None, when passed to `parser.parse_args`
```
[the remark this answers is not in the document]
benjamincarlin: argparse type makes the start position and length both type ints and as a result the python bool logic still holds. I have (as your suggestion below) moved this code into its own function that documents both the position and length as type ints

src/sctools/platform.py
```diff
@@ -629,3 +629,313 @@ def attach_barcodes(cls, args=None):
         cls._tag_bamfile(args.u2, args.output_bamfile, tag_generators)
 
         return 0
+
+
+class BarcodePlatform(GenericPlatform):
+    """Command Line Interface for extracting and attaching barcodes with specified positions
```
benjamincarlin: For an assay description: I was concerned since the class has been designed for a more generic use case, providing a particular assay as an example might confuse the user

src/sctools/platform.py
benjamincarlin: I was told to not alter the TenXV2 class until there is better test coverage.

benjamincarlin:
@ambrosejcarr Thanks for your review! I left two responses to your comments. Please let me know if there is anything else I can do!

src/sctools/platform.py
ambrosejcarr: Got it, I was confused. See comment below!

src/sctools/platform.py
```diff
@@ -629,3 +629,313 @@ def attach_barcodes(cls, args=None):
...
+
+    This class defines several methods that are created as CLI tools when sctools is installed
```
ambrosejcarr: This is where I would document the purpose of the class. 👍

ambrosejcarr:
Hi @benjamincarlin 

> ## Purpose
> Please explain the purpose of this PR and include links to any GitHub issues that it fixes:

This section is the place to put the rationale behind the PR. I apologize -- I misinterpreted the purpose of your changes and based them on our previous conversations about a specific assay. 

Are you hoping that this would provide a general solution across assays? If so, it would be a good idea to write down the assumptions the tool is making. For example, that the sample barcode would always be found on the i7 file, that the cell and molecule barcodes would be found on r1, and that no barcode is contained on u2. 

To make this a bit more general, I might omit the numerical assignments to the reads. `r1` and `i1` could become a list of one or more input fastq files, and the tool could enable a user to specify which fastq file the tag should be associated with. I make these suggestions from past experience with In-drop and Mars-seq, where the barcodes are not always in the same positions or on the same fastq files. 

I think these changes, combined with what you've already written, would do a great job of making this more general. We could talk later to @rexwangcc or @samanehsan about deprecating the 10x-specific tool in our workflows. 

Do these suggestions make sense to you?

benjamincarlin:
Hi @ambrosejcarr 

I provided a purpose and description to both the class and input file arguments. In addition, I believe that r1 and i1 becoming a list of one or more input fastq's, is expanding the scope of this PR. In order to unblock my workflow would you be willing to create an issue linking back this PR instead?

src/sctools/platform.py
```diff
@@ -629,3 +629,318 @@ def attach_barcodes(cls, args=None):
...
         return 0
+
+
+class BarcodePlatform(GenericPlatform):
+    """Command Line Interface for extracting and attaching barcodes with specified positions
+     generalizing TenXV2 attach barcodes
+
+    Sample, cell and/or molecule barcodes can be extracted and attached to an unmapped bam when the
+    corresponding barcode's start position and and length are provided. The sample barcode is extracted
+    from the index i7 fastq file and the cell/molecule barcode is extracted from the r1 fastq file
```
ambrosejcarr:
```suggestion
    from the index i7 fastq file and the cell and molecule barcodes are extracted from the r1 fastq file
```

benjamincarlin: I will create an issue and link back to this PR. Thank you for the review!

PR #66, opened 2019-05-14
AttachBarcodes generic entrypoint -- fix bugs & add test

mlin:


src/sctools/platform.py
```diff
@@ -935,15 +935,16 @@ def _make_tag_generators(
         barcode_args = {'fastq_files': r1}
 
         if i1:
-            barcode_args['embedded_barcodes'] = [cls.sample_barcode]
-            tag_generators.append(fastq.EmbeddedBarcodeGenerator(**barcode_args))
+            sample_barcode_args = dict(barcode_args)
```
ambrosejcarr: Can you explain why this change was needed in the PR description? Otherwise I agree with the tests and their assertions.

src/sctools/test/test_entrypoints.py
```diff
@@ -86,6 +86,61 @@ def test_Attach10XBarcodes_entrypoint_with_whitelist():
...
+
+    return_call = platform.BarcodePlatform.attach_barcodes(args)
+    assert return_call == 0
+    success = False
+    with pysam.AlignmentFile('test_tagged_bam.bam', 'rb', check_sq=False) as f:
+        for alignment in f:
+            if alignment.has_tag(consts.CELL_BARCODE_TAG_KEY):
+                success = True
+            # each alignment should now have a tag, and that tag should be a string
+            assert isinstance(alignment.get_tag(consts.RAW_CELL_BARCODE_TAG_KEY), str)
```
ambrosejcarr: Any interest in writing a small test that checks that the method is producing the _correct_ cell, molecule, and sample barcodes?

src/sctools/platform.py
mlin: OK

src/sctools/test/test_entrypoints.py
mlin: Will follow up time permitting, thanks

PR #69, opened 2020-01-15
snRNA-seq support in counting

barkasn:
### Purpose

### Changes

- No changes.

### Review Instructions

- No instructions.

### DIFF
--- a/src/sctools/barcode.py
+++ b/src/sctools/barcode.py
@@ -90,12 +90,12 @@ class Barcodes:
             distances.append(TwoBit.hamming_distance(a, b))
 
         keys: Tuple = (
-            'minimum',
-            '25th percentile',
-            'median',
-            '75th percentile',
-            'maximum',
-            'average',
+            "minimum",
+            "25th percentile",
+            "median",
+            "75th percentile",
+            "maximum",
+            "average",
         )
         values: List = list(np.percentile(distances, [0, 25, 50, 75, 100]))
         values.append(np.mean(distances))
@@ -185,7 +185,7 @@ class Barcodes:
 
         """
         tbe = TwoBit(barcode_length)
-        with open(file_, 'rb') as f:
+        with open(file_, "rb") as f:
             return cls(
                 Counter(tbe.encode(barcode[:-1]) for barcode in f), barcode_length
             )
@@ -277,7 +277,7 @@ class ErrorsToCorrectBarcodesMap:
         if not isinstance(errors_to_barcodes, Mapping):
             raise TypeError(
                 f'The argument "errors_to_barcodes" must be a mapping of erroneous barcodes to correct '
-                f'barcodes, not {type(errors_to_barcodes)}'
+                f"barcodes, not {type(errors_to_barcodes)}"
             )
         self._map = errors_to_barcodes
 
@@ -309,7 +309,7 @@ class ErrorsToCorrectBarcodesMap:
 
     @staticmethod
     def _prepare_single_base_error_hash_table(
-        barcodes: Iterable[str]
+        barcodes: Iterable[str],
     ) -> Mapping[str, str]:
         """Generate a map of correct barcodes and single base error codes to whitelist barcodes
 
@@ -328,7 +328,7 @@ class ErrorsToCorrectBarcodesMap:
 
             # include all single-base errors
             for i, nucleotide in enumerate(barcode):
-                errors = set('ACGTN')
+                errors = set("ACGTN")
                 errors.discard(nucleotide)
                 for e in errors:
                     error_map[barcode[:i] + e + barcode[i + 1 :]] = barcode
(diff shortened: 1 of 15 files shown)


src/sctools/count.py
```diff
@@ -90,6 +90,41 @@ def _get_alignments_grouped_by_query_name_generator(
                 )
                 yield query_name, cell_barcode, molecule_barcode, alignments
 
+    """Looks through a list of gene location to fine the one that the given read_start ovelaps
```
jessicaway:
```suggestion
    """Looks through a list of gene location to find the one that the given read_start ovelaps
```

src/sctools/test/test_groups.py
```diff
@@ -26,232 +26,272 @@ def check_parsed_metrics_csv(file_name, cell_id, class_name, expected_metrics):
...
-    with open('output_picard_group_error_summary_metrics.csv') as f:
+    with open("output_picard_group_error_summary_metrics.csv") as f:
         reader = csv.DictReader(f)
+
+        i = 0
+        match_list = []
         for line in reader:
-            assert line in expected_metrics
-    os.remove('output_picard_group_error_summary_metrics.csv')
+            for ordered_dict in expected_metrics:
```
jessicaway: We should make it clear that this test no longer considers the order of the Ordered Dictionary objects

jessicaway: If we are ok with ignoring the order of the metrics in the file, you can check the ordered dictionary from the file against a non-ordered dictionary from expected using 'in' and the order will not be considered. We do not need the extra method to compare ordered dictionaries

src/sctools/count.py
```diff
@@ -90,6 +90,41 @@ def _get_alignments_grouped_by_query_name_generator(
...
+
+    Parameters
+    ----------
+    gene_locations: Array
+        array with gene start end locations and names
+    search_start:
+        index of gene to start searching form
```
jessicaway:
```suggestion
        index of gene to start searching from
```

src/sctools/count.py
```diff
@@ -136,6 +172,9 @@ def from_sorted_tagged_bam(
         bam_file : str
             input bam file marked by cell barcode, molecule barcode, and gene ID tags sorted in that
             order
+        gene_locations : List[tuple]
```
jessicaway: Is this correct?

src/sctools/count.py
```diff
@@ -100,11 +135,12 @@ def from_sorted_tagged_bam(
         cls,
         bam_file: str,
         gene_name_to_index: Dict[str, int],
+        gene_locations: List[tuple] = None,
```
jessicaway: Incorrect? Should be a dictionary

src/sctools/count.py
```diff
@@ -90,6 +90,41 @@ def _get_alignments_grouped_by_query_name_generator(
...
+        name of gene with overlap or None if no overlap is found
+
+    """
+
+    @classmethod
+    def binary_overlap(cls, gene_locations, search_start, search_end, read_start):
+        while search_start <= search_end:
+            current_gene_index = int((search_start + search_end) / 2)
+            if (
+                gene_locations[current_gene_index][0][0]
```
jessicaway: Please rename variable for readability

src/sctools/count.py
kishorikonwar: Fixed

src/sctools/count.py
kishorikonwar: Fixed! It is a dictionary

src/sctools/count.py
kishorikonwar: added and _ to distinguish this hidden internal variable in accordance with PEP8 convention

src/sctools/test/test_groups.py
kishorikonwar: changed to unordered

src/sctools/barcode.py
```diff
@@ -277,7 +277,7 @@ def __init__(self, errors_to_barcodes: Mapping[str, str]):
         if not isinstance(errors_to_barcodes, Mapping):
             raise TypeError(
                 f'The argument "errors_to_barcodes" must be a mapping of erroneous barcodes to correct '
```
barkasn: Why isn't this changed?

rexwangcc: Weird, I was originally thinking since the single-quote -> double-quote conversion in this PR was done by `black`, which seems to intend to leave `f-string`s since they are indeed not strings but expressions. But I'm not sure as we see line 280 was converted ... Might be a bug of black and we could manually convert this...

src/sctools/test/test_entrypoints.py
```diff
@@ -282,9 +282,8 @@ def test_verify_bam_sort_raises_error_on_unsorted():
         consts.MOLECULE_BARCODE_TAG_KEY,
     ]
 
-    with pytest.raises(bam.SortError) as e:
+    with pytest.raises(bam.SortError):
         platform.GenericPlatform.verify_bam_sort(args)
-    assert 'are not in correct order' in str(e)
```
rexwangcc: This is a very good catch!!! Thank you!

PR #72, opened 2020-02-20
Fixed versions and move install of dependencies to requirements.txt

Fab-T:
### Purpose

### Changes

- Updating requirements with fixed versions
- Updated Dockerfile to run dependencies

### Review Instructions

- No instructions.


Fab-T:
Here is a PR to move all dependencies into requirements.txt so they are all scanned by Snyk and also aligning with fixed versions ( picked from the one defined in Skylab repo ).

PR #74, opened 2020-03-18
ignore multi gene alignements

barkasn:
Ignore reads that have gene name set to a list of comma separated items


barkasn:
Unit tests would be great! Kishori should be able to provide the input data (he wrote the code change I just made the PR). I would really like @kishori82 input here to make sure this is as intended.

PR #76, opened 2020-04-27
Kmk mt counts

kishorikonwar:
### Purpose
To add the number of mitochondrial genes for each cell

### Changes
 a. Added a  new option to pass the gtf file to the CalculateCellMetrics
 b. Added a function in gtf to make a set of the mitochondrial genes from the gtf file
 c. In the gathering step of cell metrics each gene name in a cell was checked if it one of the mitochondrial genes in the list and counted appropriately.

### Review Instructions
 a. Review the code change
 b. Run the pytest

### DIFF
--- a/src/sctools/gtf.py
+++ b/src/sctools/gtf.py
@@ -18,7 +18,8 @@ https://useast.ensembl.org/info/website/upload/gff.html
 
 import logging
 import string
-from typing import List, Dict, Generator, Iterable, Union
+import re
+from typing import List, Dict, Generator, Iterable, Union, Set
 
 from . import reader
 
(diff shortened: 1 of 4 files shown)


src/sctools/gtf.py
```diff
@@ -260,6 +261,46 @@ def _resolve_multiple_gene_names(gene_name: str):
...
+        gene_name = record.get_attribute("gene_name")
+        gene_id = record.get_attribute("gene_id")
+
+        if gene_name is None:
+            raise ValueError(
+                f"Malformed GTF file detected. Record is of type gene but does not have a "
+                f'"gene_name" field: {record}'
+            )
+        if re.match('^mt-', gene_name, re.IGNORECASE):
+            if gene_id not in mitochondrial_gene_ids:
```
jessicaway, line 298: Do we need this explicit check? Since this is a set, you should be able to add and entry twice without duplication.

khajoue2:
cython package is required to run the test and it is not currently in the requirements.txt file.

pytest fails. 
FAILED src/sctools/test/test_groups.py::test_write_aggregated_qc_metrics - TypeError: concat() got an unexpected keyword argument 'join_axes'

The function write_aggregated_qc_metrics has the following line:
df = pd.concat([df, dat], axis=1, join_axes=[dat.index])

I think the problem is that join_axes is deprecated in the new version of pandas.

PR #77, opened 2020-05-19
added mt molecules and percent

kishorikonwar:
### Purpose
For each cell we add the following metrics:
- n_mitochondrial_genes: the number of mitochondrial genes
- n_mitochondrial_molecules: the number of molecules from mitochondrial genes, i.e., sum of the counts from mitochondrial genes 
- pct_mitochondrial_molecules:  percentage of n_mitochondrial_molecules in terms of the total number of molecules for the cell across all genes
### Changes
modified:   src/sctools/metrics/aggregator.py
- No changes.

### DIFF
--- a/src/sctools/metrics/aggregator.py
+++ b/src/sctools/metrics/aggregator.py
@@ -415,7 +415,7 @@ class CellMetrics(MetricAggregator):
     n_mitochondrial_genes: int
         The number of mitochondrial genes detected by this cell
     n_mitochondrial_molecules: int
-        The number of molecutes from mitochondrial genes detected by this cell
+        The number of molecules from mitochondrial genes detected for this cell
     pct_mitochondrial_molecules: int
         The percentage of molecutes from mitochondrial genes detected by this cell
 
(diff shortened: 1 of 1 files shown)


src/sctools/metrics/aggregator.py
```diff
@@ -414,6 +414,10 @@ class CellMetrics(MetricAggregator):
         The number of genes that are observed by more than one read in this cell
     n_mitochondrial_genes: int
         The number of mitochondrial genes detected by this cell
+    n_mitochondrial_molecules: int
+        The number of molecutes from mitochondrial genes detected by this cell
```
khajoue2, line 418: molecutes, That's a cute misspelling!

src/sctools/metrics/aggregator.py
```diff
@@ -414,6 +414,10 @@ class CellMetrics(MetricAggregator):
...
+    pct_mitochondrial_molecules: int
+        The percentage of molecutes from mitochondrial genes detected by this cell
```
khajoue2, line 420: molecutes

PR #78, opened 2020-07-20
changed the intermediate bam file writing of the SplitBam command Sam

kishori82:
In the SplitBamByCellBarcode the intermediate files were in BAM format. In this PR we change them to SAM format for faster writing.


kishori82:
Will create the same request with a separate github accont

PR #84, opened 2020-09-30
gl_1148

nikellepetrillo:
### Purpose

### Changes

update pandas package
### Review Instructions

- No instructions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 crimson==0.5.2
-pandas==0.25.3
+pandas==1.1.1
 pysam==0.16.0.1
 pytest-cov==2.5.1
 pytest==5.1.1


pshapiro4broad:
The ticket (GL-1148)'s description says 

> There are a few places where old notations are used to call pandas, we should update this to the formatting of the latest version and update that dependency.

But the ACs only mention updating the version, not changing the old notations to new notations. If that work isn't done as part of this ticket, is there another ticket for doing this work?

khajoue2 (approved):

nikellepetrillo:
@pshapiro4broad there was a specific place farzaneh/koshori mentioned about there being a problem with the old notation. I updated it to the new notation here: https://github.com/HumanCellAtlas/sctools/pull/81/files 

I will check and see if there are any other places where there is old notation

PR #92, opened 2022-01-03
added option to output-format as either FASTQ or BAM

kishorikonwar:
### Purpose
STARSolo runs into a RAM limit for very large (over 100 GB in .fastq.gz format)  FASTQ files for 10x samples. Therefore, the solution is to partition the FASTQ files by the barcodes with a user-specified size limit. We already had a function that partitions the FASTQ sequences into BAM files. This PR modifies the C++ code to allow FASTQ options.

### Changes
We added user-specified options --output-format, which can take the options BAM or FASTQ for the split files.


khajoue2 (approved):

wjdingman1:
good to merge!

PR #103, opened 2022-08-20
Overhaul concurrency and memory management in htslib_tagsort.cpp

fredlas:
Replaces semaphore-based concurrency with a single mutex guarding the body of a single function. Also replaces a subtle memory management mechanism.

### DIFF
--- a/fastqpreprocessing/src/htslib_tagsort.cpp
+++ b/fastqpreprocessing/src/htslib_tagsort.cpp
@@ -24,35 +24,7 @@ constexpr int kThreshold = 30; // qual score threshold
 #include <stdlib.h>
 #include <unordered_map>
 #include <memory>
-#include <semaphore.h>
 #include <set>
-
-sem_t g_tagsort_semaphore;
-std::mutex g_mtx;
-std::set<unsigned int> g_threads_to_join;
-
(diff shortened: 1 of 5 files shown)


fastqpreprocessing/src/htslib_tagsort.cpp
```diff
@@ -326,41 +298,158 @@ void write_out_partial_txt_file(std::vector<TAGTUPLE> const& tuple_records,
...
+
+  // Blocks until it's this thread's turn to read, then reads a batch of alignments.
+  // Returns a pointer to the alignment ptr array, and number of alignment ptrs in the array.
+  std::pair<bam1_t**, unsigned int> readAlignments(int thread_index)
+  {
+    const std::lock_guard<std::mutex> lock(mutex_);
+    unsigned int cur_num_read = 0;
+    while (cur_num_read < options_.alignments_per_batch)
+    {
+      if (sam_read1(sam_file_ptr_, bam_hdr_, aln_arr_[thread_index][cur_num_read]) == 0)
```
fredlas, line 342: The existing code treats sam_read1() > 0 as a success, but everything I could find about the library seemed to indicate == 0 means success. We should figure out which is correct.

khajoue2 (approved):

PR #105, opened 2022-09-05
Fixes to compile sctools without patches

khajoue2:
* Fixed a memory leak in libStatGen and forked it to a new repo with all patches applied: https://github.com/khajoue2/libStatGen/releases/tag/v1.0.15.broad
* Fixed bugs in Makefile


Dockerfile
```diff
@@ -10,33 +10,27 @@ RUN mkdir /sctools/
 
 COPY . /sctools 
 
-ARG libStatGen_version="1.0.14"
 ARG htslib_version="1.13"
 
 RUN cd /sctools/fastqpreprocessing &&\
-    wget https://github.com/statgen/libStatGen/archive/v${libStatGen_version}.tar.gz &&\
+    wget https://github.com/khajoue2/libStatGen/archive/refs/tags/v1.0.15.broad.tar.gz &&\
```
jessicaway, line 16: Is this fork publicly visible? Can we make a PR to the original code base?

jessicaway (approved):
I have slight concerns about maintaining a separate fork, but for how often we update the versions of the tools we use, this will _probably_ be OK (:/)

Dockerfile
khajoue2: Yes, it is visible. The real issue is the patches and I don't think they want those changes!

